PR #1, opened 2019-07-16
add makefile and vendor

deads2k:
add makefile and vendor folder.


sttts:
/hold

until we have disable publishing, and CI + image is near to be in place

PR #2, opened 2019-07-17
Fix vendor

tnozicka:
/cc @sttts @deads2k


deads2k:
/lgtm

will merge on good images and unit

deads2k:
Image test is good. We don't use these yet.

PR #4, opened 2019-07-17
make local dockerfile consistent

deads2k:
peer to #3


deads2k:
images worked. merging

PR #6, opened 2019-07-19
Quickfix: revive legacyscheme

sttts:
There are event recorders (template, apps, maybe others) which use the encoder from the legacy scheme. This is wrong: since openshift-apiserver is separate the types are not in that scheme anymore and should not be.

This PR is a quickfix against the CI fires. It should be reverted and all references to legacy scheme should be removed.

### DIFF
--- a/cmd/openshift-controller-manager/main.go
+++ b/cmd/openshift-controller-manager/main.go
@@ -10,15 +10,41 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


cmd/openshift-controller-manager/main.go
```diff
@@ -10,16 +10,42 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
```
tnozicka: nit: \n

sttts: later

deads2k:
I think the DC failures are probably because the first set of oc images from openshift/oc got promoted.  Merging this to move us a little further forward.

PR #8, opened 2019-07-22
Remove all kube patches

deads2k:
The openshift-controller-manager has no need for kube patches.  This switches everything to straight upstream dependencies.

The binary name change is no longer needed because we no longer have a shared binary that appears the same in metrics.


mfojtik:
/lgtm

👍

PR #11, opened 2019-07-30
Bug 1730722: Add blocked registries to registries.conf

nalind:
When generating registries.conf, add the list of blocked registries from config.Spec.RegistrySources.BlockedRegistries to the list of registries in the [registries.block] section.

### DIFF
--- a/pkg/build/controller/build/build_controller.go
+++ b/pkg/build/controller/build/build_controller.go
@@ -1975,6 +1975,9 @@ func (bc *BuildController) createBuildRegistriesConfigData(config *configv1.Imag
 			Insecure: registryList{
 				Registries: registriesConfig.InsecureRegistries,
 			},
+			Block: registryList{
+				Registries: registriesConfig.BlockedRegistries,
+			},
 		},
 	}
 


nalind:
This is related to https://bugzilla.redhat.com/show_bug.cgi?id=1730722, but our plan for it (if I'm reading https://github.com/containers/image/issues/548 right) is to key off of `policy.conf` in the openshift-docker/sti-builder.

adambkaplan:
@nalind correct - in fact this has already been implemented in the build controller.

Is the gap that `policy.json` doesn't apply for image push?

nalind:
> @nalind correct - in fact this has already been implemented in the build controller.
> 
> Is the gap that `policy.json` doesn't apply for image push?

Yes.  Still working on that one.

adambkaplan:
@nalind so adding this would be a "quick fix", so to speak, and would only apply to blocked registries. It wouldn't address the whitelist case.

nalind:
> @nalind so adding this would be a "quick fix", so to speak, and would only apply to blocked registries. It wouldn't address the whitelist case.

It would probably fix the blacklisting case, but even after the policy-based changes go in I'd want us to continue supplying this information in registries.conf.

adambkaplan:
/retitle Bug 1730722: Add blocked registries to registries.conf

pkg/build/controller/build/build_controller.go
```diff
@@ -1975,6 +1975,9 @@ func (bc *BuildController) createBuildRegistriesConfigData(config *configv1.Imag
 			Insecure: registryList{
 				Registries: registriesConfig.InsecureRegistries,
 			},
+			Block: registryList{
+				Registries: registriesConfig.BlockedRegistries,
+			},
```
adambkaplan: @nalind needs at minimum a unit test - I believe I already have coverage here.

Ideally we should update our e2e suite for blocked registries in origin, too.

adambkaplan:
/lgtm

/hold

I'm going to write the PR adding the test verifying this change.

adambkaplan:
@gabemontero note that we're adding blocked registries to `registries.conf` here - I believe this aligns with what MCO is doing.

gabemontero:
it does in fact align @adambkaplan though it will be using the new formats proscribed by the MCO integration 

See https://github.com/openshift/openshift-controller-manager/pull/19/files#diff-4349d2367514c13a1d5c76dc4debf58fR2177-R2186 for example in the current prototype, though hopefully will replace that all the copied MCO code to a new MCO method being introduced via https://github.com/openshift/machine-config-operator/pull/1087

adambkaplan:
/lgtm cancel

Actually found a few issues - will open a separate PR.

adambkaplan:
/close

Replaced by #21

PR #14, opened 2019-08-12
Set BUILD_REGISTRY_SOURCES to the cluster config

nalind:
Set $BUILD_REGISTRY_SOURCES in the build pod to a JSON-encoded copy of the cluster image config's Registry Sources structure, so that the builder can enforce blocked/allowed registry lists without getting tangled up in the existing use cases for `policy.json` and `registries.conf`.


nalind:
This aims to solve https://bugzilla.redhat.com/show_bug.cgi?id=1713389, but is not tagged by it because this is opened against the master branch.

PR #15, opened 2019-08-13
Bug 1741133: Fix DC adoption

tnozicka:
/cc @soltysh 

one piece of https://github.com/openshift/origin/pull/22683

/hold

needs to be manually tested with new e2e in https://github.com/openshift/origin/pull/22683


tnozicka:
verified manually and by manually running e2e from https://github.com/openshift/origin/pull/22683

PR #19, opened 2019-08-24
Bug 1745192: configure builds to use mirrors in disconnected enivronments

gabemontero:
@openshift/openshift-team-developer-experience @bparees @nalind @mtrmac 

/assign @adambkaplan 
/assign @nalind 
/assign @mtrmac 

initially marked as WIP because I'm a little unclear on how to fully migrate the build controller's existing representation of `registries.conf` to use the V2 `registries.conf` format, but I have enough in place I believe to solicit feedback/guidance

Ultimately I'll re-enable / add more unit tests once that migration/mapping is sorted out


gabemontero:
Hmm ... doesn't like my combination of WIP and Bug ....

/hold

pkg/build/controller/build/build_controller.go
```diff
@@ -124,6 +127,25 @@ type registries struct {
 
 type tomlConfig struct {
 	Registries registries `toml:"registries"`
+}*/
+
+type endpoint struct {
```
mtrmac: Why not import the actual types from GitHub.com/containers/image, now that they have been made public for MCO doing this?

pkg/build/controller/build/build_controller.go
```diff
@@ -1978,7 +2014,15 @@ func (bc *BuildController) readClusterImageConfig() []error {
 		bc.setAdditionalTrustedCAs(additionalCAs)
 	}
 
-	registriesTOML, regErr := bc.createBuildRegistriesConfigData(imageConfig)
+	imageContentSourcePolicy, err := bc.imageContentSourcePolicyLister.Get("cluster")
```
mtrmac: There may be many ICSP objects, including multiple ICSP objects specifying the same `mirror.Source` which need to be combined as documented.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,25 +2067,41 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
 	return additionalCA.Data, nil
 }
 
-func (bc *BuildController) createBuildRegistriesConfigData(config *configv1.Image) (string, error) {
+func (bc *BuildController) createBuildRegistriesConfigData(config *configv1.Image, policy *operatorv1alpha1.ImageContentSourcePolicy) (string, error) {
 	registriesConfig := config.Spec.RegistrySources
-	if len(registriesConfig.InsecureRegistries) == 0 {
+	if len(registriesConfig.InsecureRegistries) == 0 && (policy == nil || len(policy.Spec.RepositoryDigestMirrors) == 0) {
```
mtrmac: I can’t see how not building the configuration is correct, if the code below sets `docker.io` in the search path.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,25 +2067,41 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
+	mirrors := []operatorv1alpha1.RepositoryDigestMirrors{}
+	if policy != nil {
+		mirrors = policy.Spec.RepositoryDigestMirrors
+	}
+	regs := []registry{}
+	defaultReg := registry{}
+	// docker.io must be the only entry in the registry search list
+	// See https://github.com/openshift/builder/pull/40
+	defaultReg.Location = "docker.io"
+	regs = append(regs, defaultReg)
```
mtrmac: The search list has an explicit field in the configuration; this is not that.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,25 +2067,41 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
+	for _, regStr := range registriesConfig.InsecureRegistries {
+		reg := registry{}
+		reg.Location = regStr
+		reg.Insecure = true
+		regs = append(regs, reg)
+	}
+	for _, mirror := range mirrors {
+		reg := registry{}
+		reg.Prefix = mirror.Source
```
mtrmac: Not setting `Location` is an invalid config.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,25 +2067,41 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
+		reg.Mirrors = []endpoint{}
```
mtrmac: `reg` is not used/appended anywhere?

In addition, the configuration in `registries.conf` must not contain an entry for the same `Location` twice, so merely appending does not work.

pkg/build/controller/build/build_controller_test.go
```diff
@@ -1555,14 +1558,20 @@ func TestHandleControllerConfig(t *testing.T) {
 				t.Errorf("unexpected error decoding registries config: %v", err)
```
mtrmac: I would also strongly recommend a test using the actual parser in c/image:
```go
_, err = sysregistriesv2.GetRegistries(&types.SystemContext{
	SystemRegistriesConfPath: temporaryFileWithTheGeneratedContents,
})
// err must be nil
```

pkg/build/controller/build/build_controller.go
gabemontero: Yep I looked into that but that would require a vendor/deps bump, and the github.com/containers/image vendored into this repo does not have those setting yet.

I did not want to take that on yet with a WIP / is this even the right path sort of initial change.

I should have left a TODO explaining all that.  I'll do so on the next bump.

pkg/build/controller/build/build_controller.go
gabemontero: Interesting ... @adambkaplan 's https://bugzilla.redhat.com/show_bug.cgi?id=1745192#c1 implied there would be a cluster wide object.

We'll need to reconcile that.  But yeah, if there are many vs. a singleton, that will impact the controller's informer flow, with separate event handlers for the ICSP objects, and probably a separate queue as well.

pkg/build/controller/build/build_controller.go
gabemontero: Yeah ... following the previous logic.  I'll have to defer to @bparees or @adambkaplan on the history / rationale here.

gabemontero:
@mtrmac it seems to me that the https://github.com/openshift/machine-config-operator/blob/cfb78728a6e985fc40a3ef318ba1de359de17a6d/pkg/controller/container-runtime-config/helpers.go#L321 code should be moved to https://github.com/openshift/library-go to facilitate the openshift conventions around sharing utilities like this.  In general we don't want the openshift ocm-o depending on the mco.  Certainly it could be added to library-go first, and then the mco could be subsequently changed to leverage it.

pkg/build/controller/build/build_controller.go
gabemontero: Of course the switch to https://github.com/openshift/machine-config-operator/blob/cfb78728a6e985fc40a3ef318ba1de359de17a6d/pkg/controller/container-runtime-config/helpers.go#L321 might mitigate that ... I see it is using toml

pkg/build/controller/build/build_controller.go
gabemontero: I'm exploring the MCO now using `updateRegistriesConfig` as a starting point to fully see how it fetches and munges the imageconfig and icsp's  .... based on what I'm seeing so far, don't think I'll have to divert from the current controller informer flow like I said yesterday.

gabemontero:
OK, while certainly not complete from a coding or unit-test perspective, and some specific questions noted with TODO in the implementation, I've updated the PR to hopefully progress the evaluation and discovery of next steps.

In particular:
- I'm pulling in containers/image, and leverage the systemregistries type directly
- I've copied a variety of MCO logic directly into build controller (don't want to vendor in the MCO, at least not yet ... especially if we can move some logic to library-go, especially once the TODO's are resolved) per @mtrmac 's initial pointer

pkg/build/controller/build/build_controller.go
```diff
@@ -1978,7 +2000,15 @@ func (bc *BuildController) readClusterImageConfig() []error {
 		bc.setAdditionalTrustedCAs(additionalCAs)
 	}
 
-	registriesTOML, regErr := bc.createBuildRegistriesConfigData(imageConfig)
+	imageContentSourcePolicies, err := bc.imageContentSourcePolicyLister.List(labels.Everything())
+	if err != nil && !errors.IsNotFound(err) {
+		if !errors.IsNotFound(err) {
```
mtrmac: This is going to be always true AFAICS

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,36 +2053,168 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
+	insecureRegs := []string{}
+	if config != nil {
+		insecureRegs = config.Spec.RegistrySources.InsecureRegistries
+		blockedRegs = config.Spec.RegistrySources.BlockedRegistries
+	}
+
+	//TODO the MCO code factors in the registry used for the install payload from the CVO,
+	//but seems perhaps unnecessary for builds ... in general, getting the mco and cvo objs
+	//was not a req articulated in the bug, but the `updateRegistriesConfig` I got pointed
+	//to needs data from those sources ... would rather avoid pulling those in if possible
```
mtrmac: Is this referring to `getValidRegistries` making it impossible to block `ClusterVersionStatus.Desired.Image`?

If so, I agree that this check does not need to be performed. MCO checks for that so that it’s not possible to break the cluster by setting the cluster-wide blocked registries configuration, which would be disastrous; an ordinary user breaking a single build by configuring it with too strict registry block is local and easily remedied.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,36 +2053,168 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
-			Search: registryList{
-				Registries: []string{"docker.io"},
-			},
-			Insecure: registryList{
-				Registries: registriesConfig.InsecureRegistries,
-			},
-		},
+	//TODO do not know how to reconcile this with the new type
+	// docker.io must be the only entry in the registry search list
+	// See https://github.com/openshift/builder/pull/40
```
mtrmac: Something like
```go
configObj.UnqualifiedSearchRegistries = []string{"docker.io"}
```

pkg/build/controller/build/build_controller.go
```diff
@@ -112,7 +119,7 @@ func (q *resourceTriggerQueue) Pop(key string) []string {
 	return resources
 }
 
-type registryList struct {
+/*type registryList struct {
```
mtrmac: Remove this instead of commenting out?

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,36 +2053,167 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
-			// See https://github.com/openshift/builder/pull/40
-			Search: registryList{
-				Registries: []string{"docker.io"},
-			},
-			Insecure: registryList{
-				Registries: registriesConfig.InsecureRegistries,
-			},
-		},
+
+	//TODO the MCO code factors in the registry used for the install payload from the CVO,
```
mtrmac: This is now resolved I think.

mtrmac:
> @mtrmac - on library-go, I revisited the README to make sure I understood what was involved and now remember the "high bar" warnings about what is required for inclusion.
> 
> If possible you might want to run the idea by the folks in the OWNERS file and see how amenable they would be before you go too far down this path.

In the meantime, I have created https://github.com/openshift/machine-config-operator/pull/1087 . Please check that the proposed API works well.

mtrmac:
In addition to the MCO PR linked above, I have also filed https://github.com/openshift/library-go/pull/513 for the library-go inclusion proposal.

mtrmac:
> Not sure why `verify-deps` fails in CI but passes for me locally .... looking into the `vendor/github.com/spf13/cobra/cobra/tpl` package it is complaining about ... the bump in container/images must somehow be messing with how this is vendored in ... might explore temporarily clearing out the vendor tree and then running `glide install ...` ... or maybe move things around in glide.yaml

The files it’s complaining about really exist upstream, and really are not included in the commit.

I’ve recently run into similar difficulties caused by a `.gitignore` in one of the vendor repos excluding some of the files — locally the output of `glide` contained both the `.gitignore` all of those ignored, but `git add` / `git commit -a` silently ignored them; they were available locally, so local checks passed for me, but they were not a part of the commit, and CI checks failed. That does not seem to be quite the situation here, just something to look into.

gabemontero:
clean test runs

mtrmac:
I’d still much prefer the code to exist in one place, but I’m not sure whether we can make that happen in time.

In the meantime, could you apply https://github.com/openshift/machine-config-operator/pull/1082 to this copy, please?

gabemontero:
at this point @mtrmac I'd rather see https://github.com/openshift/machine-config-operator/pull/1087  merge and I refactor towards that ... if the library-go attempt gets traction, we can pivot.

before we push on that though, let me revisit the "do we need this in 4.2" thread with the various parties I mentioned in your MCO PR and based on the results, go from there

gabemontero:
I've sent out an email to the interested parties to get a decision on the risk of merging https://github.com/openshift/machine-config-operator/pull/1087 for 4.2 and then basing this PR off of that

adambkaplan:
@gabemontero https://github.com/openshift/machine-config-operator/pull/1087 is now LGTM - we'll likely want to copy the code in with the same `registries` subpackage.

mtrmac:
Not copy, vendor in and call, please.

gabemontero:
OK MCO is vendored in, calling @mtrmac 's new API, and rebased on recent changes from @adambkaplan 

@adambkaplan bump / PTAL

glide.lock
```diff
@@ -14,18 +14,27 @@ imports:
...
 - name: github.com/certifi/gocertifi
   version: ee1a9a0726d2ae45f54118cac878c990d4016ded
+- name: github.com/containerd/continuity
+  version: 75bee3e2ccb6402e3a986ab8bd3b17003fc0fdec
+  subpackages:
+  - pathdriver
 - name: github.com/containers/image
-  version: 4bc6d24282b115f8b61a6d08470ed42ac7c91392
-  repo: https://github.com/openshift/containers-image.git
+  version: abf32c4ea589cb8e96bdca5a478dba68f11980e5
```
mtrmac: This is v2.0.0; please use at least v3.0.0.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,39 +2040,42 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
-	if len(registriesConfig.InsecureRegistries) == 0 && len(registriesConfig.BlockedRegistries) == 0 {
+func (bc *BuildController) createBuildRegistriesConfigData(config *configv1.Image, policies []*operatorv1alpha1.ImageContentSourcePolicy) (string, error) {
+
+	blockedRegs := []string{}
+	insecureRegs := []string{}
+	if config != nil {
+		insecureRegs = config.Spec.RegistrySources.InsecureRegistries
+		blockedRegs = config.Spec.RegistrySources.BlockedRegistries
+	}
+	if len(insecureRegs) == 0 && len(blockedRegs) == 0 {
```
mtrmac: (I know I’ve pointed at this line before…) This will cause ICSP to be ignored (= disconnected not to work) if this branch is taken.

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,39 +2040,42 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
-	if len(configTOML) == 0 {
-		klog.V(4).Info("using default registry settings for builds")
+
+	var newData bytes.Buffer
+	encoder := toml.NewEncoder(&newData)
+	if err := encoder.Encode(configObj); err != nil {
+		return "", err
+	}
+
+	if len(newData.Bytes()) == 0 {
```
mtrmac: I can’t see how this can happen (at least the `UnqualifiedSearchRegistries` entry should be present), and anyway, the effect is the same as the default `return string(…)` below, apart from the log entries.

glide.lock
gabemontero: @mtrmac We don't specify a containers/image version in `glide.yaml`

This version in `glide.lock` comes from the versions we do specify, which in this case is machine-config-operator.

And if you look there:  https://github.com/openshift/machine-config-operator/blob/master/Gopkg.toml#L168-L170 and https://github.com/openshift/machine-config-operator/blob/release-4.2/Gopkg.toml

It specifies v2.0.0.0:

```
[[constraint]]
  name = "github.com/containers/image"
  version = "v2.0.0"
```

You want use to use v3.0.0.0, then update machine-config-operator accordingly and we can refresh our dependency once that is done.

I do not want to get into the business of specifying containers/image if we are not coding to it specifically.  And you'll need to give me a good, precise reason why we have to run at a different verison of that vs. machine-config-operator.

pkg/build/controller/build/build_controller.go
gabemontero: yes you did .... some stuff got lost in the various rebases ... will address in next push

gabemontero: I do know I had a version as some point that had ICSP factored in here

pkg/build/controller/build/build_controller.go
gabemontero: What if the `mco.EditRegistriesConfig` API we now call after a future dependency update decided to empty out our setting of the search registries ....

Leaving this debug in "just in case"

pkg/build/controller/build/build_controller.go
gabemontero: ok pushed the update for this ... now `if len(insecureRegs) == 0 && len(blockedRegs) == 0 && len(policies) == 0 {`

gabemontero:
ok update pushed for @mtrmac 's https://github.com/openshift/openshift-controller-manager/pull/19#discussion_r319996771

Rationales for not making adjustments for the other 2 posted.

@adambkaplan - PTAL in general as well as my rationale for @mtrmac 's other 2 comments 

My thought is to merge this as is if we are OK with the code changes, and handle an MCO bump to get a containers/image bump separately if needed after QE verification kicks in.

pkg/build/controller/build/build_controller.go
mtrmac: That’s a fair point — and there already is another code path with `return "", nil`, so even that is not exceptional.

glide.lock
gabemontero: @mtrmac and I talked offline in slack ... we've agreed to live with the current version of containers/image in this PR.

I'll queue up a separate discussion with @adambkaplan and @bparees the pros and cons for possible explicit citation of containers/image in the OCM repo as part of a subsequent change.

pkg/build/controller/build/build_controller.go
gabemontero: ok cool ... we'll leave this as is then - thanks

gabemontero:
and we have passing test @adambkaplan ... bump

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,39 +2040,42 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
+		klog.V(4).Info("using default insecure registry settings for builds")
 		return "", nil
 	}
 	klog.V(4).Info("overrode registry settings for builds")
-	klog.V(5).Infof("generated registries.conf for build pods: \n%s", string(configTOML))
-	return string(configTOML), nil
+	klog.V(0).Infof("generated registries.conf for build pods: \n%s", string(newData.Bytes()))
```
adambkaplan: I would not expect the full `registries.conf` file to be dumped into normal logs. Please revert to V(5)

pkg/build/controller/build/build_controller.go
```diff
@@ -2023,39 +2040,42 @@ func (bc *BuildController) getAdditionalTrustedCAData(config *configv1.Image) (m
...
-	}
 
-	configTOML, err := toml.Marshal(configObj)
+	configObj := sysregistriesv2.V2RegistriesConf{}
+	// docker.io must be the only entry in the registry search list
+	// See https://github.com/openshift/builder/pull/40
+	configObj.UnqualifiedSearchRegistries = []string{"docker.io"}
+	err := mco.EditRegistriesConfig(&configObj, insecureRegs, blockedRegs, policies)
 	if err != nil {
+		klog.V(4).Infof("MCO library had problem building registries config: %s", err.Error())
```
adambkaplan: Use a higher log level for an error - `V(0)`?

gabemontero:
On Wed, Sep 4, 2019 at 10:55 AM Adam Kaplan <notifications@github.com>
wrote:

> *@adambkaplan* requested changes on this pull request.
>
> @gabemontero <https://github.com/gabemontero> nits on log levels.
> otherwise looks good.
>
> For 4.3 I'd like to see us create a library-go-esque repo to help us
> generate the essential container runtime config files, since these are used
> across MCO, builds, and image stream import:
>
yeah Miloslav and I have talked about that for 4.3
 As his library-go PR did not get much traction

pkg/build/controller/build/build_controller.go
gabemontero: ah missed that as part of debug/dev ... thanks part of next push

pkg/build/controller/build/build_controller.go
gabemontero: sure switched to v0

gabemontero:
updates pushed @adambkaplan ...PTAL

PR #24, opened 2019-09-05
Bug 1749557: new containers/image requires context with docker reference NewImageS…

gabemontero:
…ource

more stringent context requirements (new remote calls added) for download image signatures with new containers/image from recent bump

/assign @bparees 
/assign @adambkaplan 
@openshift/openshift-team-developer-experience FYI / PTAL

### DIFF
--- a/pkg/image/controller/signature/container_image_downloader.go
+++ b/pkg/image/controller/signature/container_image_downloader.go
@@ -36,7 +36,11 @@ func (s *containerImageSignatureDownloader) DownloadImageSignatures(image *image
 	if err != nil {
 		return nil, err
 	}
-	source, err := reference.NewImageSource(nil, nil)
+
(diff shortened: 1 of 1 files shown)


bparees:
there's an implication here that we don't have test coverage for image signature import or we would have caught this when the dep bump was done.  is it a test that we lost when we lost the integration tests?  can we bring it back as part of this PR?

gabemontero:
@dmage or @adambkaplan would have to answer the integration test question @bparees  ... I know I re-introduced a bunch of integration tests and extended tests late July / early August, but they were build focused.

gabemontero:
One set of passing tests ... the e2e-aws shows green, but had one DC flake with post commit hooks that seems unrelated 

also going to run builds and image eco

adambkaplan:
/retitle Bug 1749557: Provide context when obtaining image signatures

PR #25, opened 2019-09-18
remove all controllers except  quota and security

sanchezl:


bparees:
I thought the plan was to remove apiserver team owned controllers and leave the devex owned controllers here..   where are these going?

PR #26, opened 2019-09-18
remove all controllers except  quota and security (master)

sanchezl:


deads2k:
/hold 

we don't want to merge, but we do want to see it buidl. Great.

PR #29, opened 2019-09-25
Bug 1755115: do not schedule imports for streams with sha/digest references whose …

gabemontero:
…import has completed

per instructions from @smarterclayton  ... the api needs to be  “level driven”

with this change, we can use scheduled import to enact retry for imagestreams with dockerimage refs that container the canonical sha/digest, but once they are successful, no longer try to import, as the sha/digest is the final outcome

/assign @bparees 
/assign @adambkaplan 
/assign @dmage

### DIFF
--- a/pkg/image/controller/scheduled_image_controller.go
+++ b/pkg/image/controller/scheduled_image_controller.go
@@ -13,6 +13,7 @@ import (
 
 	imagev1 "github.com/openshift/api/image/v1"
 	imagev1lister "github.com/openshift/client-go/image/listers/image/v1"
+	imageref "github.com/openshift/library-go/pkg/image/reference"
 	metrics "github.com/openshift/openshift-controller-manager/pkg/image/metrics/prometheus"
 )
 
@@ -179,6 +180,13 @@ func resetScheduledTags(stream *imagev1.ImageStream) {
 	next := stream.Generation + 1
 	for tag, tagRef := range stream.Spec.Tags {
 		if tagImportable(tagRef) && tagRef.ImportPolicy.Scheduled {
+			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
+				continue
+			} else if len(ref.ID) > 0 && !tagNeedsImport(stream, tagRef, true) {
+				// ref.ID is set if this is a canonical, sha/digest ref;
+				// we allow scheduled import of those, but only until the initially succeed
+				continue
+			}
 			tagRef.Generation = &next
 			stream.Spec.Tags[tag] = tagRef
 		}
@@ -189,6 +197,13 @@ func resetScheduledTags(stream *imagev1.ImageStream) {
 func needsScheduling(stream *imagev1.ImageStream) bool {
 	for _, tagRef := range stream.Spec.Tags {
 		if tagImportable(tagRef) && tagRef.ImportPolicy.Scheduled {
+			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
+				return false
+			} else if len(ref.ID) > 0 && !tagNeedsImport(stream, tagRef, true) {
+				// ref.ID is set if this is a canonical, sha/digest ref;
+				// we allow scheduled import of those, but only until the initially succeed
+				return false
+			}
 			return true
 		}
 	}
--- a/pkg/image/controller/scheduled_image_controller_test.go
+++ b/pkg/image/controller/scheduled_image_controller_test.go
@@ -124,3 +124,131 @@ func TestScheduledImport(t *testing.T) {
 		t.Fatalf("should have not added item to scheduler: %#v", sched.scheduler)
 	}
 }
+
+func TestSchedulingChecks(t *testing.T) {
+	one := int64(1)
+	tests := map[string]struct {
+		is             *imagev1.ImageStream
+		expectedResult bool
+	}{
+		"none scheduled": {
+			is: &imagev1.ImageStream{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test", Namespace: "other", UID: "1", ResourceVersion: "1",
+					Annotations: map[string]string{imagev1.DockerImageRepositoryCheckAnnotation: "done"},
+					Generation:  1,
+				},
+				Spec: imagev1.ImageStreamSpec{
+					Tags: []imagev1.TagReference{
+						{
+							Name:         "default",
+							From:         &corev1.ObjectReference{Kind: "DockerImage", Name: "mysql:latest"},
+							Generation:   &one,
+							ImportPolicy: imagev1.TagImportPolicy{},
+						},
+					},
+				},
+				Status: imagev1.ImageStreamStatus{
+					Tags: []imagev1.NamedTagEventList{
+						{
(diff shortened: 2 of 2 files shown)


gabemontero:
this change does not require it, but users will ultimately need https://github.com/openshift/openshift-apiserver/pull/28 merged to be able to leverage this new path

pkg/image/controller/scheduled_image_controller.go
```diff
@@ -189,6 +190,13 @@ func resetScheduledTags(stream *imagev1.ImageStream) {
 func needsScheduling(stream *imagev1.ImageStream) bool {
 	for _, tagRef := range stream.Spec.Tags {
 		if tagImportable(tagRef) && tagRef.ImportPolicy.Scheduled {
+			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
+				return false
```
bparees: shouldn't the validation code prevent this from happening?  (it's the same check the validation is doing, right?)

bparees: i guess if you have to parse it anyway to get the id it makes sense to respect the error, if any though.

pkg/image/controller/scheduled_image_controller_test.go
```diff
@@ -124,3 +124,124 @@ func TestScheduledImport(t *testing.T) {
...
+					Name: "test", Namespace: "other", UID: "1", ResourceVersion: "1",
+					Annotations: map[string]string{imagev1.DockerImageRepositoryCheckAnnotation: "done"},
+					Generation:  1,
+				},
+				Spec: imagev1.ImageStreamSpec{
+					Tags: []imagev1.TagReference{
+						{
+							Name:         "default",
+							From:         &corev1.ObjectReference{Kind: "DockerImage", Name: "abc@sha256:3c87c572822935df60f0f5d3665bd376841a7fcfeb806b5f212de6a00e9a7b25"},
+							Generation:   &one,
```
bparees: this test would be better if the spec generation was newer than the status generation, to confirm we still don't import a SHA tag that is already successfully imported, right?

bparees:
i think there are some issues here....

needsScheduling looks at the imagestream as a whole to make a decision, so if any tag within it needs to be scheduled, the whole imagestream will be scheduled.

Then when the schedule actually comes up, resetScheduledTags is called to setup the specific tags for re-import.  So resetScheduledTags needs to also avoid resetting SHA tags that are already imported.  (probably needs to make the same needsImport check)

pkg/image/controller/scheduled_image_controller_test.go
bparees: ok on further study of the code, these values matching is our signal that the import is done/successful.  But i'm still wondering how the spec.generation value gets updated...i'm wondering if someone updates the imagestream, whether spec.generation gets updated and that triggers another round of imports.  If we can confirm that's not the case, then i think it's reasonable to assume that "once imported successfully, spec.generation will never exceed spec.status, for a sha based tag, as long as our schedule import avoidance logic is sound"

pkg/image/controller/scheduled_image_controller.go
gabemontero: that is where I landed on this one as well

pkg/image/controller/scheduled_image_controller_test.go
gabemontero: Ok it looks like the logic which says a tagRef has changed are here:  https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/registry/imagestream/strategy.go#L375-L423

That logic above then dictates whether the tagRef.generation is updated at https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/registry/imagestream/strategy.go#L217-L228

I believe all the conditions at https://github.com/openshift/openshift-apiserver/blob/master/pkg/image/apiserver/registry/imagestream/strategy.go#L375-L423 would be valid reasons for re-importing in our case, but let me know if you disagree @bparees

gabemontero:
@bparees I've pushed an update to address the problem you found in https://github.com/openshift/openshift-controller-manager/pull/29#issuecomment-535502573

pkg/image/controller/scheduled_image_controller_test.go
bparees: yup, i agree, thanks for digging into it @gabemontero

pkg/image/controller/scheduled_image_controller.go
```diff
@@ -180,6 +180,13 @@ func resetScheduledTags(stream *imagev1.ImageStream) {
 	next := stream.Generation + 1
 	for tag, tagRef := range stream.Spec.Tags {
 		if tagImportable(tagRef) && tagRef.ImportPolicy.Scheduled {
+			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
+				continue
+			} else if len(ref.ID) > 0 && !tagNeedsImport(stream, tagRef, true) {
+				// ref.ID is set if this is a canonical, sha/digest ref;
+				// we allow scheduled import of those, but only until the initially succeed
+				continue
+			}
```
bparees: i think it is still reasonable to also do this check in needsScheduling.  (if no tags in the imagestream end up needing scheduling, we save ourselves some work)

pkg/image/controller/scheduled_image_controller_test.go
gabemontero: cool thanks

pkg/image/controller/scheduled_image_controller.go
gabemontero: ok cool yeah that is why I left it there

bparees: oh, i must have only been looking at the new commit, so i thought you moved it to here only.  nm.

bparees:
squash + lgtm
/approve

gabemontero:
squashed @bparees

bparees:
/lgtm

let's make sure this(and the related changes) gets some good soak time before backporting to 4.2 though

pkg/image/controller/scheduled_image_controller.go
```diff
@@ -189,6 +197,13 @@ func resetScheduledTags(stream *imagev1.ImageStream) {
 func needsScheduling(stream *imagev1.ImageStream) bool {
 	for _, tagRef := range stream.Spec.Tags {
 		if tagImportable(tagRef) && tagRef.ImportPolicy.Scheduled {
+			if ref, err := imageref.Parse(tagRef.From.Name); err != nil && len(tagRef.From.Name) > 0 {
+				return false
+			} else if len(ref.ID) > 0 && !tagNeedsImport(stream, tagRef, true) {
+				// ref.ID is set if this is a canonical, sha/digest ref;
+				// we allow scheduled import of those, but only until the initially succeed
+				return false
```
dmage: What if we have two tags and the second one needs to be scheduled?

gabemontero: good catch @dmage 

I'll queue up a follow up PR momentarily.

bparees: yes, thanks @dmage, this should be a continue like is in resetScheduledTags.  (same for the err check above)

PR #30, opened 2019-09-27
describes how to set up a local dev env

p0lyn0mial:


hack/dev_env.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+# symlink the current kube config to $DEV_DIR_NAME - it will be set in config.json later
+if [[ -z "${KUBECONFIG}" ]]; then
+  echo "KUBECONFIG env var is required for this script to work, please set it and run the script again"
+  exit 1
+fi
+
+# create a directory where all files will be stored
+DEV_DIR_NAME="dev_env"
```
p0lyn0mial: we could move it into `hack` dir

pkg/cmd/openshift-controller-manager/openshiftcontrolplane_default.go
```diff
@@ -3,14 +3,23 @@ package openshift_controller_manager
...
 
 func setRecommendedOpenShiftControllerConfigDefaults(config *openshiftcontrolplanev1.OpenShiftControllerManagerConfig) {
-	configdefaults.SetRecommendedHTTPServingInfoDefaults(config.ServingInfo)
+	if config.ServingInfo == nil {
+		config.ServingInfo = &configv1.HTTPServingInfo{}
+		configdefaults.SetRecommendedHTTPServingInfoDefaults(config.ServingInfo)
+	} else if config.ServingInfo != nil && len(config.ServingInfo.BindAddress) == 0 {
+		klog.Warning("config.ServingInfo will be ignored as it contains an empty BindAddress")
+		config.ServingInfo = nil
+	}
```
sttts: what is the goal of this block? When will either case happen?

p0lyn0mial: the idea was to make `config.ServingInfo` optional so that OCM doesn't expose `health` EP over SSL as this would require less step to run it on a local machine.

the first condition will be satisfied when there no `"servingInfo"` entry is in the configuration file
the second condition will be satisfied when there is empty `"servingInfo": {}` entry in the config file

sttts: can you add a godoc one-liner for each?

p0lyn0mial: done - https://github.com/openshift/openshift-controller-manager/pull/30/commits/e6329e8c1f7618519b670ee96d33c09870b8c281

pkg/cmd/openshift-controller-manager/openshiftcontrolplane_default.go
```diff
@@ -3,14 +3,25 @@ package openshift_controller_manager
...
 )
 
 func setRecommendedOpenShiftControllerConfigDefaults(config *openshiftcontrolplanev1.OpenShiftControllerManagerConfig) {
-	configdefaults.SetRecommendedHTTPServingInfoDefaults(config.ServingInfo)
+	if config.ServingInfo == nil {
+		// this condition  will be satisfied when there no "servingInfo" entry is in the configuration file
+		config.ServingInfo = &configv1.HTTPServingInfo{}
+		configdefaults.SetRecommendedHTTPServingInfoDefaults(config.ServingInfo)
+	} else if config.ServingInfo != nil && len(config.ServingInfo.BindAddress) == 0 {
+		// this condition will be satisfied when there is empty "servingInfo": {} entry in the configuration file
```
sttts: I meant more high level: `// disable serving if already set with empty bind address, used in dev setup`

p0lyn0mial: okay, updated - https://github.com/openshift/openshift-controller-manager/pull/30/commits/56c7e6f3d62986879b56bb416ecf3e3ad606359b

PR #33, opened 2019-10-02
add leader lock for cluster-policy-controller upgrades

sallyom:
I've added a leader lock in the NamespaceSCCAllocationController for upgrades 4.2->4.3
cluster-policy-controller container is split from openshift-controller-manager in 4.3 and will run in kube-controller-manager static pod

@deads2k


deads2k:
required to remove bootstrapping cycle.  tagging.

PR #34, opened 2019-10-11
Bug 1753731: seeding of img registry hostname with config if needed

gabemontero:
/assign @bparees - some additional questions came to my mind after coding the changes we talked about today

### DIFF
--- a/pkg/build/controller/build/build_controller.go
+++ b/pkg/build/controller/build/build_controller.go
@@ -966,7 +966,18 @@ func (bc *BuildController) resolveImageStreamTag(ref *corev1.ObjectReference, li
 		}
 		return nil, fmt.Errorf("the referenced image stream %s/%s could not be found: %v", namespace, name, err)
 	}
-	if newRef, ok := imageutil.ResolveLatestTaggedImage(stream, tag); ok {
+	newRef := ""
+	ok = false
+	if len(stream.Status.DockerImageRepository) > 0 {
+		newRef, ok = imageutil.ResolveLatestTaggedImage(stream, tag)
+	} else {
+		streamCopy := stream.DeepCopy()
+		// in case the api server has not yet picked up the internal registry hostname from the cluster wide
+		// OCM config, we use our copy here to facilitate leveraging pull through with local tag reference policy
+		streamCopy.Status.DockerImageRepository = bc.internalRegistryHostname
+		newRef, ok = imageutil.ResolveLatestTaggedImage(streamCopy, tag)
+	}
+	if ok {
 		// generate informational event if
 		// a) tag has spec
 		// b) tag is local reference


pkg/build/controller/build/build_controller.go
```diff
@@ -966,23 +966,54 @@ func (bc *BuildController) resolveImageStreamTag(ref *corev1.ObjectReference, li
...
+			if len(stream.Status.DockerImageRepository) == 0 {
+				bc.recorder.Eventf(build, corev1.EventTypeNormal,
+					"LocalRefImageStreamTagUnavailableNoImageRegistryRef",
+					"The build %s/%s references an image stream tag %s with the local reference policy type, but the internal image registry location was not marked in the image stream",
+					build.Namespace,
+					build.Name,
+					ref.Name)
+			}
+			latestTaggedImage := imageutil.LatestTaggedImage(stream, tag)
+			if len(latestTaggedImage.Image) == 0 {
```
bparees: i'm not understanding the purpose of this...  if the import hasn't happened/succeeded yet, won't `imageutil.ResolveLatestTaggedImage(streamCopy, tag)` have already returned you nil/an error?

bparees: https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L247

pkg/build/controller/build/build_controller.go
```diff
@@ -966,23 +966,54 @@ func (bc *BuildController) resolveImageStreamTag(ref *corev1.ObjectReference, li
...
+				//TODO len(latestTaggedImage.Image) == 0 can also result in not using the IST
+				// local tag reference policy;
+				// see https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L297
+				// see https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L304
+				// see https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L333
+				// if stream.Status.DockerImageRepository
+				// is empty, how likely is it that the Image field will be set?
+				// should we do something "crazy" like poll/wait on the stream??
```
bparees: no.  it's valid to define an imagestream that uses localref policy, and not have an internal registry configured.

in such a case, i'd expect imagestream resolution to return the external registry spec.  Whether we can then pull that spec or not is unknown, but it's certainly *possible* if the user has the right creds in place.

Anyway that's the behavior we had implemented (if not local registry exists, use the external registry spec when resolving the imagestream) so i think we should preserve it.

pkg/build/controller/build/build_controller.go
```diff
@@ -966,23 +966,54 @@ func (bc *BuildController) resolveImageStreamTag(ref *corev1.ObjectReference, li
...
+				// should build controller and image controller have some sort of synchronization,
+				// given they are in the same process, where the build controller puts an event back on the queue
+				// (though for a limited number of retries) if a required imagestream is not ready yet ?
+				bc.recorder.Eventf(build, corev1.EventTypeNormal,
```
bparees: see above, i don't think we need this, i don't think we get here if the imagestreamtag couldn't be resolved to an imported status.tag, and we already handle that error flow up the chain.

pkg/build/controller/build/build_controller.go
gabemontero: I'm worried about this conditional at https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L333

Some of the surrounding context around that `if`:

```
	case imagev1.LocalTagReferencePolicy:
		local := stream.Status.DockerImageRepository
		if len(local) == 0 || len(latest.Image) == 0 {
			// fallback to the originating reference if no local docker registry defined or we
			// lack an image ID
			return latest.DockerImageReference
		}
```

Do we think the `|| len(latest.Image) == 0` is unnecessary there?

If not, then my additional alert test tells us if we had a situation where `len(stream.Status.DockerImageRepository)` indicated the img reg was set, but since `latest.Image` was not populated, we still did not set up local pullthrough.

pkg/build/controller/build/build_controller.go
gabemontero: OK, as a reminder though, what is biting us in the extended tests is that the external registry spec is to the TBR, and our tests do not have the TBR creds included in the build's pull secret.

Pending the clarification on https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L333 , and if we can't further mitigate the "current imagestream entry in the lister cache is stale, etc.", perhaps we should start looking at a utility in the extended tests to mine the TBR creds from the install pull secret and seeding it as a pull secret in our extended test build pull secrets, just in case.

pkg/build/controller/build/build_controller.go
gabemontero: responded to your above ... please see and let's reconcile.

pkg/build/controller/build/build_controller.go
bparees: > Do we think the || len(latest.Image) == 0 is unnecessary there?

i think it's unnecessary in your specific flow.  It may be useful to other callers.  

you are calling this:
https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L293

which calls this:
https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L238

which checks this and returns nil if there are no valid entries associated w/ the tag:
https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L246-L247

which means when it then calls this:
https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L297

it will return this:
https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L304

which means you will never enter the if block here:
https://github.com/openshift/openshift-controller-manager/pull/34/files#diff-4349d2367514c13a1d5c76dc4debf58fR978

so you'll never get to:
https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L333

Whether that line is unnecessary in general i can't say, there might be other callers of https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L321 that come in with different arguments.

> If not, then my additional alert test tells us if we had a situation where len(stream.Status.DockerImageRepository) indicated the img reg was set, but since latest.Image was not populated, we still did not set up local pullthrough.

if latest.image is not populated, nothing else matters.  if latest.image is not populated, there is no pullspec/sha available, which is a scenario this code already handled and errors reported.  If you run a build for a buildconfig that points to an imagestream that has not successfully imported and you'll see it:

```
$ oc describe is bar
Name:			bar
Namespace:		p1
Created:		4 minutes ago
Labels:			<none>
Annotations:		openshift.io/image.dockerRepositoryCheck=2019-10-11T17:45:05Z
Image Repository:	image-registry.openshift-image-registry.svc:5000/p1/bar
Image Lookup:		local=false
Unique Images:		0
Tags:			2

latest
  tagged from docker.io/foo/bar

  ! error: Import failed (Unauthorized): you may not have access to the container image "docker.io/foo/bar:latest"
      4 minutes ago

test
  tagged from docker.io/foo/bar:test

  ! error: Import failed (Unauthorized): you may not have access to the container image "docker.io/foo/bar:test"
      18 seconds ago

$ oc start-build nodejs-ex
The ImageStreamTag "bar:test" is invalid: from: Error resolving ImageStreamTag bar:test in namespace p1: unable to find latest tagged image
```

pkg/build/controller/build/build_controller.go
bparees: > OK, as a reminder though, what is biting us in the extended tests is that the external registry spec is to the TBR, and our tests do not have the TBR creds included in the build's pull secret.

yes, understood, but the intent of your fix here is that we will properly resolve the imagestreamtag to the local reference because we are not relying on the internal registry hostname being set in the copy of the imagestream that we have, which might be stale, but instead are using the internal registry hostname value that the OCM knows about, which is not stale because the OCM gets restarted when that value becomes available.

The theoretical timing window here is:

1) internal registry hostname becomes available in the cluster
2) OCM restarts, caches imagestream values.  Because the apiserver has not restarted yet, the imagestream values the OCM gets, do not have the registry hostname decorated onto them.  The cache will never be "fixed" unless the imagestream changes again for some reason.
3) apiserver restarts to pick up internal registry hostname (decorator logic for gets is now in effect, but it's too late).
4) OCM starts a build, uses the cached imagestream that lacks the registry hostname to resolve a tag, so it resolves the tag to the external registry.

The intent of your changes is that you will not be reliant on the internal registry hostname being set on the imagestream via the decorator, which means you are not dependent on the apiserver restarting before OCM restarts.

pkg/build/controller/build/build_controller.go
gabemontero: Thanks for all the detail, that helps.

Also, I *finally* found the decoration stuff you mentioned earlier.  I've looked at it before, but could not remember where it was exactly.  It is part of etcd storage.

So bottom line, the `stream.Status.DockerImageRepository` gets set on ImageStream creation, update, and it not tied to the IST imports in any way (which is what was the core of what I was worried about when thinking if it was possible for the setting of those items at https://github.com/openshift/library-go/blob/master/pkg/image/imageutil/helpers.go#L333 to somehow be related).

I'll clean up the unneeded checks and update the PR momentarily.

gabemontero:
updates pushed @bparees PTAL

pkg/build/controller/build/build_controller.go
```diff
@@ -966,7 +966,18 @@ func (bc *BuildController) resolveImageStreamTag(ref *corev1.ObjectReference, li
...
+	ok = false
+	if len(stream.Status.DockerImageRepository) > 0 {
+		newRef, ok = imageutil.ResolveLatestTaggedImage(stream, tag)
+	} else {
+		streamCopy := stream.DeepCopy()
+		// in case the api server has not yet picked up the internal registry hostname from the cluster wide
+		// OCM config, we use our copy here to facilitate leveraging pull through with local tag reference policy
+		streamCopy.Status.DockerImageRepository = bc.internalRegistryHostname
+		newRef, ok = imageutil.ResolveLatestTaggedImage(streamCopy, tag)
+	}
```
bparees: so is this the only place where we ever resolve an imagestreamtag for a buildconfig? 

i know we potentially allow imagestreamtag references in various places in buildconfigs(e.g. image inputs in addition to builder images), hopefully they all go through here but have you sanity checked that?

gabemontero: `image_trigger_controller.go` makes a call to `imageutil.ResolveLatestTaggedImage` , and some of the builconfig image change trigger methods make calls to methods that call it.

But it seemed in those cases, the flow are driven off of imagestream events, so in theory we have an updated imagestream and should in theory have that image registry hostname set as part of getting the event.

Concur ?

Otherwise, all the build_controller flows, as part of creating the build pod, go through this method.

bparees: yeah i think that sounds ok.

gabemontero: cool thanks

PR #35, opened 2019-10-15
Fixes Bug 1760776: Fix targetPort in route created by ingress

derkoe:
Use the service port as the Route's targetPort
(the service port may differ from the service targetPort).

See also https://github.com/openshift/origin/pull/23951 and the associated issue https://github.com/openshift/origin/issues/23937

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -583,14 +583,16 @@ func targetPortForService(namespace string, path *extensionsv1beta1.HTTPIngressP
 		expect := path.Backend.ServicePort.StrVal
 		for _, port := range service.Spec.Ports {
 			if port.Name == expect {
-				return &port.TargetPort
+				targetPort := intstr.FromString(port.Name)
+				return &targetPort
 			}
 		}
 	} else {
 		for _, port := range service.Spec.Ports {
 			expect := path.Backend.ServicePort.IntVal
 			if port.Port == expect {
-				return &port.TargetPort
+				targetPort := intstr.FromInt(int(port.Port))
+				return &targetPort
 			}
 		}
 	}
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -583,16 +583,26 @@ func targetPortForService(namespace string, path *extensionsv1beta1.HTTPIngressP
 		expect := path.Backend.ServicePort.StrVal
 		for _, port := range service.Spec.Ports {
 			if port.Name == expect {
-				targetPort := intstr.FromString(port.Name)
-				return &targetPort
+				if port.TargetPort.Type == intstr.String {
+					targetPort := intstr.FromString(port.Name)
+					return &targetPort
+				}
+				return &port.TargetPort
 			}
 		}
 	} else {
 		for _, port := range service.Spec.Ports {
 			expect := path.Backend.ServicePort.IntVal
 			if port.Port == expect {
-				targetPort := intstr.FromInt(int(port.Port))
-				return &targetPort
+				if port.TargetPort.Type == intstr.String {
+					if &port.Name == nil {
+						// TODO lookup numeric port in endpoint
+						return nil
+					}
+					targetPort := intstr.FromString(port.Name)
+					return &targetPort
+				}
+				return &port.TargetPort
 			}
 		}
 	}
--- a/pkg/route/ingress/ingress_test.go
+++ b/pkg/route/ingress/ingress_test.go
@@ -204,7 +204,7 @@ func TestController_stabilizeAfterCreate(t *testing.T) {
 					{
 						Name:       "http",
 						Port:       80,
-						TargetPort: intstr.FromInt(8080),
+						TargetPort: intstr.FromString("http"),
 					},
 				},
 			},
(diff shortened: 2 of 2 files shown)


pkg/route/ingress/ingress_test.go
```diff
@@ -369,7 +369,6 @@ func TestController_sync(t *testing.T) {
 			Spec: v1.ServiceSpec{
 				Ports: []v1.ServicePort{
 					{
-						Name:       "80-tcp",
```
danehans: @derkoe what is the purpose for removing the name from this port?

derkoe:
Tests seem to fail because Terraform cannot provision the cluster:
> level=fatal msg="failed to fetch Cluster: failed to generate asset \"Cluster\": failed to create cluster: failed to apply using Terraform"

Try again
/retest

Miciah:
I'm not listed as an owner, but
/lgtm

danehans:
`timed out waiting for the condition`

/test e2e-aws

danehans:
/retitle Fixes Bug 1760776: Fix targetPort in route created by ingress

ironcladlou:
/hold

I perceive some confusion around this issue. Please keep the hold until @smarterclayton has weighed in.

pkg/route/ingress/ingress.go
```diff
@@ -583,14 +583,16 @@ func targetPortForService(namespace string, path *extensionsv1beta1.HTTPIngressP
...
+				return &targetPort
 			}
 		}
 	} else {
 		for _, port := range service.Spec.Ports {
 			expect := path.Backend.ServicePort.IntVal
 			if port.Port == expect {
-				return &port.TargetPort
+				targetPort := intstr.FromInt(int(port.Port))
+				return &targetPort
```
Miciah: See https://github.com/openshift/origin/pull/23951/files#r337271800.

derkoe: I have addressed this issue in 5113ff5.

This changes the int port handling back to the way it was and only addresses the named port issues.

There are two cases for named Pod ports:
1. the Service has a named port which references a named port in the Pod. This is resolved by my commit.
2. the Service has an unnamed port which references a named port in the Pod. **This is still open** - I have added a test case for that `"create route - targetPort string, service port has no name"` but the implementation is a bit too much for me. In this case the port has to read from the corresponding Endpoint and set in the Route.

pkg/route/ingress/ingress.go
```diff
@@ -583,13 +583,25 @@ func targetPortForService(namespace string, path *extensionsv1beta1.HTTPIngressP
...
 				return &port.TargetPort
 			}
 		}
 	} else {
 		for _, port := range service.Spec.Ports {
 			expect := path.Backend.ServicePort.IntVal
 			if port.Port == expect {
+				if port.TargetPort.Type == intstr.String {
+					if &port.Name == nil {
+						// TODO lookup numeric port in endpoint
```
derkoe: @Miciah maybe you can help here - this is the special case where the Service port is unnamed but references the Pod port via name. Here the controller has to lookup the target port in the corresponding Endpoint.

Miciah: A service port must have a name if the service has more than 1 port.  If the service does not have more than 1 port, then the route's `spec.to.name` field can be left unspecified, and the route will use whatever port exists on the endpoints resource.

If a service port does have a name, then the corresponding port on the endpoints resource will have the same name, so I think it would be reasonable always to use the service port name if it exists, and otherwise omit `spec.to.name` from the route per the above reasoning.  Does that make sense?

I find this tricky to reason about, so in case it is helpful to you or others, I'll elaborate on my reasoning.  Each service port either has a name or does not, and the port specifies its target either by name or by number.  Thus we have four cases to consider:

| Service port `name` | Service port `targetPort` | Route `spec.to.name`     |
|------------------- |------------------------- |------------------------ |
| absent              | number                    | empty                    |
| absent              | name                      | empty                    |
| present             | number                    | port name or target port |
| present             | name                      | port name                |

In the first two cases, we do not have a service port name, but the service can have only 1 port (because if it had multiple ports, the ports would need to have names), so specifying a value for the route's `spec.to.name` field would be superfluous anyway.

In the third case, the service may have multiple ports (thus we need to set the route's `spec.to.name` field), we have a service port name and a target port number, and either one could be used to specify the route's `spec.to.name` field. I suggest using the service port name for consistency with the fourth case (below), but I could accept the argument that it would be preferable to use the target port number for consistency with the current logic.

In the fourth case, the service may have multiple ports (same as the third case), and we have a port name but no target port number, so we must use the port name for the route's `spec.to.name`. (Note that this is the case that is broken with the current code, because the current code uses the target port name, which may differ from the port name on the endpoints resource.)

Miciah:
@derkoe, did https://github.com/openshift/openshift-controller-manager/pull/35#discussion_r338405851 make sense?  Are you able to continue work on this change, or should I take over?  Thanks for your efforts thus far!

derkoe:
@Miciah yes it would be great you could take over - I think this change goes to deep for me

derkoe:
@Miciah should I  close this PR?`

Miciah:
Yeah, this PR is superseded by #53 and can be closed.  Thanks again for your earlier efforts!

/close

PR #39, opened 2019-10-23
Bug 1758014: Mitigation for policy passing bug.

ricardomaraschini:
Mitigates a bug on passing policies to build pod by inspecting the
namespace looking for preexisting ConfigMaps and aborting if any is
found. Builds who fell under such situation are tagged as Failed and
the reason informs that we are failed because the namespace is
"tainted".

### DIFF
--- a/pkg/build/controller/build/build_controller.go
+++ b/pkg/build/controller/build/build_controller.go
@@ -1208,25 +1208,65 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
 		return update, nil
 	}
 
-	klog.V(4).Infof("Pod %s/%s for build %s is about to be created", build.Namespace, buildPod.Name, buildDesc(build))
-	pod, err := bc.podClient.Pods(build.Namespace).Create(buildPod)
-	if err != nil && !errors.IsAlreadyExists(err) {
-		// Log an event if the pod is not created (most likely due to quota denial).
-		bc.recorder.Eventf(build, corev1.EventTypeWarning, "FailedCreate", "Error creating build pod: %v", err)
-		update.setReason(buildv1.StatusReasonCannotCreateBuildPod)
-		update.setMessage("Failed creating build pod.")
-		return update, fmt.Errorf("failed to create build pod: %v", err)
+	existingPod, err := bc.podClient.Pods(build.Namespace).Get(
+		buildPod.Name, metav1.GetOptions{},
+	)
+	if err != nil && !errors.IsNotFound(err) {
+		update.setReason(buildv1.StatusReasonError)
+		update.setMessage("Failed getting build pod.")
+		return update, fmt.Errorf("failed to get build pod: %v", err)
+	} else if errors.IsNotFound(err) {
+		// Before starting a new pod we need to make sure the namespace
+		// doesnt contain pre filled in ConfigMaps as these may contain
+		// invalid content.
+		if tainted, err := bc.taintedNamespace(build); err != nil {
+			update.setReason(buildv1.StatusReasonError)
+			update.setMessage("Failed verifying namespace.")
+			return update, fmt.Errorf("failed verifying namespace: %v", err)
+		} else if tainted {
+			update = transitionToPhase(
+				buildv1.BuildPhaseFailed,
+				buildv1.StatusReasonError,
+				"Tainted environment",
+			)
+			return update, fmt.Errorf("tainted environment")
+		}
 
-	} else if err != nil {
-		bc.recorder.Eventf(build, corev1.EventTypeWarning, "FailedCreate", "Pod already exists: %s/%s", buildPod.Namespace, buildPod.Name)
-		klog.V(4).Infof("Build pod %s/%s for build %s already exists", build.Namespace, buildPod.Name, buildDesc(build))
+		klog.V(4).Infof(
+			"Pod %s/%s for build %s is about to be created",
+			build.Namespace, buildPod.Name, buildDesc(build),
+		)
 
-		// If the existing pod was not created by this build, switch to the
-		// Error state.
-		existingPod, err := bc.podClient.Pods(build.Namespace).Get(buildPod.Name, metav1.GetOptions{})
+		pod, err := bc.podClient.Pods(build.Namespace).Create(buildPod)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/build/controller/build/build_controller.go
```diff
@@ -1872,6 +1893,39 @@ func (bc *BuildController) findOwnedConfigMap(owner *corev1.Pod, namespace strin
...
+		buildutil.GetBuildCAConfigMapName(build),
+		buildutil.GetBuildGlobalCAConfigMapName(build),
+		buildutil.GetBuildSystemConfigMapName(build),
+	} {
+		if hasCM, err := bc.findConfigMap(
+			build.Namespace, name,
+		); err != nil {
+			return false, err
+		} else if hasCM {
+			return true, nil
```
adambkaplan: I think this can return a false positive. Consider the following scenario:

1. Build is created
2. Build controller creates the build pod
3. Build controller creates the additional CA configMap
4. Build controller gets an apiserver error creating the global trust CA configMap

@bparees based on my understanding, in this scenario the build controller will re-sync and try to continue setting up the build pod. With this PR the build will fail permanently because one of the ConfigMaps already exists.

pkg/build/controller/build/build_controller.go
```diff
@@ -1208,25 +1208,65 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
...
+		update.setReason(buildv1.StatusReasonError)
+		update.setMessage("Failed getting build pod.")
+		return update, fmt.Errorf("failed to get build pod: %v", err)
+	} else if errors.IsNotFound(err) {
+		// Before starting a new pod we need to make sure the namespace
+		// doesnt contain pre filled in ConfigMaps as these may contain
+		// invalid content.
+		if tainted, err := bc.taintedNamespace(build); err != nil {
+			update.setReason(buildv1.StatusReasonError)
+			update.setMessage("Failed verifying namespace.")
```
adambkaplan: We'll need a clearer error message - such as "Unexpected build configuration data found."

pkg/build/controller/build/build_controller.go
```diff
@@ -1208,25 +1208,65 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
...
+			return update, fmt.Errorf("failed verifying namespace: %v", err)
+		} else if tainted {
+			update = transitionToPhase(
+				buildv1.BuildPhaseFailed,
+				buildv1.StatusReasonError,
+				"Tainted environment",
```
adambkaplan: Ditto - should use the same message as above.

pkg/build/controller/build/build_controller.go
```diff
@@ -1268,25 +1308,6 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
 				return update, err
 			}
 		}
-
-	} else {
```
adambkaplan: @ricardomaraschini we still need this block to handle the case where the build pod was created, but the child ConfigMaps have not all been created yet.

pkg/build/controller/build/build_controller.go
ricardomaraschini: That is not the case @adambkaplan. If the pod already exists we don't check for "tainted namespace", we just operate as we used to do prior to this PR. Is there something I am not seeing?

pkg/build/controller/build/build_controller.go
ricardomaraschini: The error message here should say something like: "I was trying to look for unexpected build configuration data but I have failed to do so, will try your build again later". Not sure how to express that properly.

"Failed looking for unexpected build configuration" ?

pkg/build/controller/build/build_controller.go
ricardomaraschini: Here your previous comment fits perfectly :-)

bparees:
This looks to me like at best it's shrinking some timing windows, not really fixing the problem.

Even if i can't create the CM up front, i can potentially edit it before the build runs, especially if i can manage to prevent the build pod from starting or being scheduled right away.

I think the correct solution here needs to:

1) create the CMs somewhere that a user can't create/edit them
2) have the build pod itself fetch the content as part of init, rather than mounting the content in from a location the user can control.

unfortunately (1) creates additional challenges, especially around garbage collection (GC can't span namespaces, as i recall).

Other options would include injecting the content into the build pod as an env var but that might get too big if someone has a ton of CAs.

might need some more brainstorming on this one.

pkg/build/controller/build/build_controller.go
ricardomaraschini: This is still there but in a different position. This branch now is executed when we know that the pod does not exist(`errors.IsNotFound(err)`) on https://github.com/openshift/openshift-controller-manager/pull/39/files#diff-4349d2367514c13a1d5c76dc4debf58fR1218

ricardomaraschini:
@bparees I have created a doc around this issue yesterday and the proposed solution there was(as you proposed) to make the `builder` to read the info directly. This one here was intended as a "mitigation" as I still don't know how aggressive the changes are going to be on the `builder` side of the force(I am starting to evaluate that). Anyways, we may close this one and go only for the `builder` patch.

gabemontero:
> With this PR I worry a bit about failure modes when creating the build pod. IIRC it's possible for the build pod to be created, but sit pending because a temporary issue prevents us from creating the child ConfigMaps.

Certainly there have been similar issues with secrets/serviceaccounts in the past.  I could certainly see something similar with the ConfigMaps we started to leverage, but I don't recall those as readily, though certainly there was the gap in initial 4.0/4.1 builds were I was not as involved.

> 
> I wonder if a more sustainable approach is to update the `createXXConfigMap` methods to do a synchronization:
> 
>     1. If CM does not exist, create it with the content we know is desired
> 
>     2. If CM does exist, inspect the owner refs and data. If not deep equal, do an update.

I suppose this would handle a user manually creating the config maps.  If the situation is some rogue older version created by the build controller, that would not handle this, but perhaps the naming is sufficient for that.

I wonder if creation timestamps could serve as another useful cross reference.  In theory the build should always be older than the config map, right?

> 
> 
> @bparees @gabemontero thoughts?

ricardomaraschini:
Following our discussions I am closing this PR and the original bugzilla. The reasoning is written as a comment on the BZ.

PR #43, opened 2019-10-29
Bug 1764498: Remove references to extensions/v1beta1.

ricardomaraschini:
This is deprecated and we should be using apps/v1 instead.


ricardomaraschini:
@adambkaplan @dmage This patch solved the issue but I need help in evaluating the impact of it. This patch affects `TemplateInstanceController`: events related to `Deployments` reported by `TemplateInstanceInformer` that fall under `extensions/v1beta1` will be ignored, i.e. no more checks for "Readiness".

PR #49, opened 2019-11-20
Bug 1774492: add push secret as pull secret with incremental builds

gabemontero:
/assign @adambkaplan 

My analysis of the logs from this bug matched what you originally came up with @adambkaplan 

Here is my take on the raw fix

Took a quick scan at the unit tests and there isn't a starting point to easily add one for this

Most likely we would want to create an extended test equivalent of the reproducer in the bug for verification

### DIFF
--- a/pkg/build/controller/strategy/sti.go
+++ b/pkg/build/controller/strategy/sti.go
@@ -192,6 +192,9 @@ func (bs *SourceBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 
 	setOwnerReference(pod, build)
 	setupDockerSecrets(pod, &pod.Spec.Containers[0], build.Spec.Output.PushSecret, strategy.PullSecret, build.Spec.Source.Images)
+	if strategy.Incremental != nil && *strategy.Incremental {
+		mountSecretVolume(pod, &pod.Spec.Containers[0], build.Spec.Output.PushSecret.Name, DockerPullSecretMountPath, "pull-incremental")
+	}
 	// For any secrets the user wants to reference from their Assemble script or Dockerfile, mount those
 	// secrets into the main container.  The main container includes logic to copy them from the mounted
 	// location into the working directory.


gabemontero:
missed the unit test piece - thanks

On the builder side, originally I thought we would not need additional changes, as daemonless just cycles through all the auths here:  https://github.com/openshift/builder/blob/master/pkg/build/builder/daemonless.go#L123-L131

But looking again, I now see it assume only one file at the mount point via https://github.com/openshift/builder/blob/master/pkg/build/builder/cmd/dockercfg/cfg.go#L104-L129

So per your point 2, we could change what we have here and toggle the pull secret based on whether incremental is in play, vs. trying to add multiple ones as I have here.

Then the `log.V(5).Infof("Setting authentication for registry %q at %q.", registry, ac.ServerAddress)` from `buildDaemonlessImage` in openshift/builder would find the right one 

thoughts?

gabemontero:
no I'm turned around ... we need to be able to mount both secrets and change builder to access both as needed

gabemontero:
I'll add the unit test, but wait on some weigh in on priority before moving to openshift/builder

gabemontero:
OK my merry-go-round continues ... forgot that I also discovered this piece in builder:  https://github.com/openshift/builder/blob/master/pkg/build/builder/sti.go#L240-L256

or 

```
	if config.Incremental {
		if s.build.Spec.Strategy.SourceStrategy.ForcePull || !isImagePresent(s.dockerClient, config.IncrementalFromTag) {
			// Per @bparees the dockercfg.PushTypeAuth is needed to use the same credentials/authentication that
			// we used to push the image previously.
			searchPaths := dockercfg.NewHelper().GetDockerAuthSearchPaths(dockercfg.PushAuthType)
			startTime := metav1.Now()
			err = s.pullImage(config.IncrementalFromTag, searchPaths)
			timing.RecordNewStep(ctx, buildapiv1.StagePullImages, buildapiv1.StepPullInputImage, startTime, metav1.Now())
			// If there was an error, the incremental image may not exist. Treat the build as a normal s2i build.
			if err != nil {
				log.V(2).Infof("Failed to pull incremental builder image %s - executing normal s2i build instead.", config.IncrementalFromTag)
				log.V(5).Infof("Incremental image pull failure: %v", err)
				config.Incremental = false
				config.IncrementalFromTag = ""
			}
		}
	}
```

There is logic to already pull the image at the start.

Still looking ... but perhaps we just turn off force pull prior to the build starting, so it does not attempt to re-pull with the wrong creds.

PR #51, opened 2019-12-04
DEVEXP-489: Rebase k8s 117

gabemontero:
/assign @adambkaplan 

@openshift/openshift-team-developer-experience fyi


gabemontero:
@soltysh @mfojtik please let me know if you want someone from your respective teams to be officially assigned to this PR

pkg/cmd/openshift-controller-manager/controller_manager.go
```diff
@@ -25,9 +25,6 @@ import (
 	origincontrollers "github.com/openshift/openshift-controller-manager/pkg/cmd/controller"
 	"github.com/openshift/openshift-controller-manager/pkg/cmd/imageformat"
 	"github.com/openshift/openshift-controller-manager/pkg/version"
-
-	// for metrics
-	_ "k8s.io/kubernetes/pkg/client/metrics/prometheus"
```
adambkaplan: @deads2k was this import used for anything? If so, is there something to replace it?

pkg/cmd/controller/apiserver_handlers.go
```diff
@@ -16,9 +18,9 @@ func newBypassAuthorizer(auth kauthorizer.Authorizer, paths ...string) kauthoriz
 	return bypassAuthorizer{paths: sets.NewString(paths...), authorizer: auth}
 }
 
-func (a bypassAuthorizer) Authorize(attributes kauthorizer.Attributes) (allowed kauthorizer.Decision, reason string, err error) {
+func (a bypassAuthorizer) Authorize(ctx context.Context, attributes kauthorizer.Attributes) (allowed kauthorizer.Decision, reason string, err error) {
```
adambkaplan: /cc @deads2k

pkg/cmd/openshift-controller-manager/controller_manager.go
gabemontero: Did a little more research on this one @adambkaplan 

that `_` import was done very early on in the ocm historywith https://github.com/openshift/openshift-controller-manager/commit/b7179b40d0ecbe386d9a8a079fc2224a5505cf62 as part of the `v0.0.0-alpha0` tag back in Nov 2018.

Today, there are absolutely no other references to that package in the repo.

And across the board now all the controllers in OCM are using `github.com/prometheus/client_golang/prometheus`

Assuming @deads2k is spending his cycles elsewhere I think we can assume this change is in the benign / cleanup category, especially given that it no longer exists in k8s/k8s.

pkg/cmd/controller/apiserver_handlers.go
gabemontero: fyi @adambkaplan this change was needed as a result of an upstream change in k8s apiserver

see https://github.com/kubernetes/apiserver/blob/kubernetes-1.17.0-rc.1/pkg/authorization/authorizer/interfaces.go#L71

pkg/cmd/openshift-controller-manager/controller_manager.go
adambkaplan: If it's no longer upstream, then I am going to assume this is a no-op. Good news is that we do have test coverage on the build metrics in our conformance suite, and this is green.

PR #56, opened 2019-12-16
Bug 1775973: use OCM config registry hostname with imagechangetrigger in case imag…

gabemontero:
…estream cache is stale (a la builds) for local ref streams

/assign @bparees 
per our recent theory


gabemontero:
metrics startup failure e2e-aws

/retest

PR #60, opened 2020-01-07
Add ingress-to-route controller OWNERS

ironcladlou:
Add the maintainers to a new OWNERS file.

### DIFF
--- a/pkg/route/OWNERS
+++ b/pkg/route/OWNERS
@@ -1,0 +1,19 @@
+reviewers:
+  - bparees
+  - smarterclayton
+  - mfojtik
+  - soltysh
+  - csrwng
+  - gabemontero
+  - adambkaplan
+  - ironcladlou
+  - Miciah
+  - danehans
+  - frobware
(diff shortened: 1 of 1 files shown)


pkg/route/OWNERS
```diff
@@ -0,0 +1,19 @@
...
+  - Miciah
+  - danehans
+  - frobware
+approvers:
+  - bparees
+  - mfojtik
+  - soltysh
+  - adambkaplan
+  - ironcladlou
+  - Miciah
```
bparees: do you really want me/adam/michal/maciej as approvers on this dir?  seems like it should be your team + group lead + architect.

bparees: (most of us are here anyway, also:
https://github.com/openshift/openshift-controller-manager/blob/master/OWNERS#L9)

ironcladlou: I went ahead and changed it to match the OWNERS of the other repos we maintain

PR #62, opened 2020-01-30
Bug 1777337: employ k8s label value validation when creating build pod build label…

gabemontero:
… value

/assign @adambkaplan

### DIFF
--- a/pkg/build/buildutil/util.go
+++ b/pkg/build/buildutil/util.go
@@ -257,8 +257,18 @@ func GetBuildSystemConfigMapName(build *buildv1.Build) string {
 // label in a pod. If the length of the string parameter exceeds
 // the maximum label length, the value will be truncated.
 func LabelValue(name string) string {
-	if len(name) <= validation.DNS1123LabelMaxLength {
-		return name
+	end := len(name)
+	newName := name
+	// first, try to truncate from the end to find a valid
+	// label
+	for end > 0 {
+		errStrs := validation.IsDNS1123Label(newName)
+		if len(errStrs) == 0 {
+			return newName
+		}
+		end--
+		newName = newName[:end]
 	}
-	return name[:validation.DNS1123LabelMaxLength]
+	klog.Warningf("In creating the value of the build label in the build pod, several attempts at manipulating %s to meet k8s label name requirements failed", name)
+	return name
 }
--- a/pkg/build/buildutil/util_test.go
+++ b/pkg/build/buildutil/util_test.go
@@ -8,6 +8,32 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+func TestLabelValue(t *testing.T) {
+	testCases := []struct {
+		name           string
+		input          string
+		expectedOutput string
+	}{
+		{
+			name:           "do-not-end-with-hyphen",
+			input:          "cluster-kube-controller-manager-operator-4.3.0.ipv6-2019-11-27-0001-build",
+			expectedOutput: "cluster-kube-controller-manager-operator-4",
+		},
+		{
+			name:           "typical-name",
+			input:          "ruby-1-build",
+			expectedOutput: "ruby-1-build",
+		},
+	}
+
+	for _, tc := range testCases {
+		result := LabelValue(tc.input)
+		if result != tc.expectedOutput {
+			t.Errorf("tc %s got %s for %s instead of %s", tc.name, result, tc.input, tc.expectedOutput)
+		}
+	}
+}
+
 func TestMergeEnvWithoutDuplicates(t *testing.T) {
 	testCases := []struct {
 		name                string


pkg/build/buildutil/util.go
```diff
@@ -42,8 +43,22 @@ const (
 	sysConfigConfigMapSuffix = "sys-config"
 	// GlobalCAConfigMapKey is the key into the config map data for the injected CA data
 	GlobalCAConfigMapKey = "ca-bundle.crt"
+
+	// this is the regexp for the k8s pod label validation
+	// unfortunately a public variable is not exposed for anything related to the regexp,
```
gabemontero: perhaps an opportunity for an upstream contribution, but IMO we don't gate our fix based on that, as per the bugzilla this came from a CI failure to build

pkg/build/buildutil/util.go
```diff
@@ -257,8 +272,36 @@ func GetBuildSystemConfigMapName(build *buildv1.Build) string {
...
+	// first, try to truncate from the end to find a valid
+	// label
+	for end > 0 {
+		if labelRegexpCompile.MatchString(newName) {
+			return newName
+		}
+		end--
+		newName = newName[:end]
+	}
+	// otherwise, try to truncate from the both ends
```
gabemontero: I'm inclined to not even deal with this flavor, since this name is suppose to be derived from the build config name, which has already gone through similar validations.

But I leave it up for now to allow for comments / opinions from others.

pkg/build/buildutil/util_test.go
```diff
@@ -8,6 +8,37 @@ import (
...
+		input          string
+		expectedOutput string
+	}{
+		{
+			name:           "do-not-end-with-hyphen",
+			input:          "cluster-kube-controller-manager-operator-4.3.0.ipv6-2019-11-27-0001-build",
+			expectedOutput: "cluster-kube-controller-manager-operator-4.3.0.ipv6-2019-11-27",
+		},
+		{
+			name:           "do-not-start-with-hyphen",
```
gabemontero: this test case would go away if the "it is based on the build config name which has already been validated, etc." notion is agreed upon.

pkg/build/buildutil/util.go
gabemontero: I found the public API in k8s that allows for us to validate

Will update PR shortly.

gabemontero:
PR updated to use public k8s validation method (which I was too dense to notice before) @adambkaplan PTAL

gabemontero:
have green e2e's @adambkaplan

pkg/build/buildutil/util.go
```diff
@@ -257,8 +272,18 @@ func GetBuildSystemConfigMapName(build *buildv1.Build) string {
...
+	for end > 0 {
+		errStrs := validation.IsDNS1123Label(newName)
+		if len(errStrs) == 0 {
+			return newName
+		}
+		end--
+		newName = newName[:end]
 	}
-	return name[:validation.DNS1123LabelMaxLength]
+	klog.Warningf("In creating the value of the build label in the build pod, several attempts at manipulating %s to meet k8s label name requirements failed", name)
```
adambkaplan: Will we ever log this - would something like an all-symbols label value fail?

pkg/build/buildutil/util_test.go
```diff
@@ -8,6 +8,32 @@ import (
...
+func TestLabelValue(t *testing.T) {
+	testCases := []struct {
+		name           string
+		input          string
+		expectedOutput string
+	}{
+		{
+			name:           "do-not-end-with-hyphen",
+			input:          "cluster-kube-controller-manager-operator-4.3.0.ipv6-2019-11-27-0001-build",
+			expectedOutput: "cluster-kube-controller-manager-operator-4",
```
adambkaplan: I don't see this test case ending with a hyphen?

pkg/build/buildutil/util.go
```diff
@@ -42,8 +43,22 @@ const (
 	sysConfigConfigMapSuffix = "sys-config"
 	// GlobalCAConfigMapKey is the key into the config map data for the injected CA data
 	GlobalCAConfigMapKey = "ca-bundle.crt"
+
+	// this is the regexp for the k8s pod label validation
+	// unfortunately a public variable is not exposed for anything related to the regexp,
+	// on the length of the label value; in fact, the regexp noted in the error message is
+	// is not the entire regexp
+	labelRegexp = "^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$"
```
adambkaplan: Regex not used - we can remove it.

pkg/build/buildutil/util.go
gabemontero: yes which is why I left it ... agree the possibility is remote, but seems more positive than negative to me

pkg/build/buildutil/util_test.go
gabemontero: Yeah this is the value from the bug ... the 63rd character is a hyphen

pkg/build/buildutil/util.go
gabemontero: ah dammit ... cleaning up in a sec

gabemontero:
original bits cleaned up @adambkaplan

gabemontero:
various unrelated flakes e2e-aws-builds

/retest

gabemontero:
tests green again @adambkaplan

PR #76, opened 2020-03-17
WIP - DEVEXP-423: Mounting node's pull secrets on build pod.

ricardomaraschini:
Mounting node pull secrets on build pod, according to:

openshift/enhancements/enhancements/node-pull-credentials


pkg/build/controller/strategy/docker.go
```diff
@@ -102,6 +107,15 @@ func (bs *DockerBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 						EmptyDir: &v1.EmptyDirVolumeSource{},
 					},
 				},
+				{
+					Name: "node-pullsecrets",
+					VolumeSource: v1.VolumeSource{
+						HostPath: &v1.HostPathVolumeSource{
```
gabemontero: So this comment here @ricardomaraschini is as much of a broader comment for builds in general as it is for this piece (though we *could* start addressing things with your changes).

Namely, we've added another thing that necessitates a privileged pod.

So, thoughts that come to mind @ricardomaraschini @adambkaplan @bparees :
- we have reqs on the books for devising a build solution that does not require privileged pods, no?
- if I'm wrong there, ignore the rest :-)
- are we hoping to address that with build v1, or will we defer to build v2 for this?
- if not build v1, ignore the rest :-)
- do we have the concise list of places we need privileged?  and if so, should this feature be added to that list?
- and finally, to facilitate that future work, should we start abstracting out privileged requiring elements like this into utility methods, decorating/mutating abstractions, that in the future could consume a config setting and if these privileged elements should be added to the build pod?

thanks

pkg/build/controller/strategy/docker.go
```diff
@@ -39,6 +39,7 @@ func (bs *DockerBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 
 	privileged := true
 	strategy := build.Spec.Strategy.DockerStrategy
+	typeFile := v1.HostPathType("File")
```
adambkaplan: Shouldn't this be a constant in the API?

pkg/build/controller/strategy/docker.go
adambkaplan: > we have reqs on the books for devising a build solution that does not require privileged pods, no?

Unprivileged builds won't happen until we drop support for RHEL7 workers. We need fuse-overlay to give us some level of performance parity.

> are we hoping to address that with build v1, or will we defer to build v2 for this?

It's unclear at this point if builds v2 will be the place where we land the unprivileged builds feature. I like this idea since it lowers the barrier for the upstream community to consume. I.e. builds v2 could be a regular workload on any k8s cluster, _but_ it works best if you have Fedora CoreOS/RHCOS as your host nodes.

> do we have the concise list of places we need privileged? and if so, should this feature be added to that list?

We're already going down the path of adding more host path dependencies with the build image cache enhancement, so I wouldn't worry about this right now.

> and finally, to facilitate that future work, should we start abstracting out privileged requiring elements like this into utility methods, decorating/mutating abstractions, that in the future could consume a config setting and if these privileged elements should be added to the build pod?

I'm not inclined to do this right now. In many respects I'd prefer builds v2 avoid host dependencies so that it could run on any k8s cluster. If there's demand for it, then we can look to refactor things out.

pkg/build/controller/strategy/docker_test.go
```diff
@@ -89,13 +89,15 @@ func TestDockerCreateBuildPod(t *testing.T) {
 	// container storage
 	// blobs content cache
 	// global CA injection configmap
-	if len(container.VolumeMounts) != 11 {
-		t.Fatalf("Expected 11 volumes in container, got %d", len(container.VolumeMounts))
+	// node pull secrets
+	if len(container.VolumeMounts) != 12 {
+		t.Fatalf("Expected 12 volumes in container, got %d", len(container.VolumeMounts))
```
adambkaplan: @ricardomaraschini Need to add the node pull secret to the source strategy (sti) and custom strategy builds, too.

pkg/build/controller/strategy/docker.go
gabemontero: > > we have reqs on the books for devising a build solution that does not require privileged pods, no?
> 
> Unprivileged builds won't happen until we drop support for RHEL7 workers. We need fuse-overlay to give us some level of performance parity.
> 
> > are we hoping to address that with build v1, or will we defer to build v2 for this?
> 
> It's unclear at this point if builds v2 will be the place where we land the unprivileged builds feature. I like this idea since it lowers the barrier for the upstream community to consume. I.e. builds v2 could be a regular workload on any k8s cluster, _but_ it works best if you have Fedora CoreOS/RHCOS as your host nodes.
> 
> > do we have the concise list of places we need privileged? and if so, should this feature be added to that list?
> 
> We're already going down the path of adding more host path dependencies with the build image cache enhancement, so I wouldn't worry about this right now.
> 
> > and finally, to facilitate that future work, should we start abstracting out privileged requiring elements like this into utility methods, decorating/mutating abstractions, that in the future could consume a config setting and if these privileged elements should be added to the build pod?
> 
> I'm not inclined to do this right now. In many respects I'd prefer builds v2 avoid host dependencies so that it could run on any k8s cluster. If there's demand for it, then we can look to refactor things out.

my last comment was meant to be in the context of build v1, not build v2.  Regardless, it only has bearing if the RHEL7/fuse_overlay items you noted initially are addressed.

pkg/build/controller/strategy/docker.go
ricardomaraschini: It turns out we do have a constant on `corev1`. I have sent over a commit to address this.

pkg/build/controller/strategy/docker_test.go
ricardomaraschini: Thanks! I forgot those. I think we have a problem when it comes to custom strategy as by mounting these secrets one may easily copy them to the outside world, no? 

Our [documentation](https://docs.openshift.com/container-platform/4.1/builds/build-strategies.html#build-strategy-custom-build_build-strategies) states that: 

"Custom builds run with a very high level of privilege and are not available to users by default. Only users who can be trusted with cluster administration permissions should be granted access to run custom builds."

What do you reckon?

bparees: good question @ricardomaraschini.  my view would be that it's reasonable to leave this out of custom builds for now and force those users to still provide their own pull secrets.  People doing custom builds already have to be pretty sophisticated.

If we get a lot of pushback about the pain that causes we can revisit the decision, since as you note, someone doing custom builds also already has to be a highly trusted user.

adambkaplan: @bparees that sounds like a fair compromise to me. We need to doc this in JIRA so we make this clear in the release notes.

PR #78, opened 2020-03-26
Revert "[release 4.3] Bug 1779282: Use OwnerRefs to clean up SA pull secrets"

smarterclayton:
Reverts openshift/openshift-controller-manager#72

Is possibly the cause of https://bugzilla.redhat.com/show_bug.cgi?id=1785023

/hold

while we triage


smarterclayton:
Real fix is in 4.5 and appears to work

PR #83, opened 2020-03-30
Bug 1801415: Migrate extensions/v1beta1 -> networking.k8s.io/v1beta1

ironcladlou:
https://bugzilla.redhat.com/show_bug.cgi?id=1801415

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -8,8 +8,8 @@ import (
 
 	"k8s.io/klog"
 
-	"k8s.io/api/core/v1"
-	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1beta1 "k8s.io/api/networking/v1beta1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -22,10 +22,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
-	extensionsinformers "k8s.io/client-go/informers/extensions/v1beta1"
+	networkingv1beta1informers "k8s.io/client-go/informers/networking/v1beta1"
 	kv1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
-	extensionslisters "k8s.io/client-go/listers/extensions/v1beta1"
+	networkingv1beta1listers "k8s.io/client-go/listers/networking/v1beta1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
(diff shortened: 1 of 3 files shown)


Miciah:
Do we need to convert objects using the old API, as https://github.com/kubernetes/ingress-nginx/pull/4127/commits/84102eec2ba270f624c57023aab59aab4471178e#diff-ece58b3bebec02a6d3173c8d8099d298R904-R939 does?

ironcladlou:
cc @deads2k @sttts @mfojtik can you help us understand what else we might need to do here (https://github.com/openshift/openshift-controller-manager/pull/83#issuecomment-606658312)?

pkg/route/ingress/ingress.go
```diff
@@ -514,7 +514,7 @@ func newRouteForIngress(
 			Labels:       ingress.Labels,
 			Annotations:  ingress.Annotations,
 			OwnerReferences: []metav1.OwnerReference{
-				{APIVersion: "extensions/v1beta1", Kind: "Ingress", Controller: &t, Name: ingress.Name, UID: ingress.UID},
+				{APIVersion: "networking.k8s.io/v1beta1", Kind: "Ingress", Controller: &t, Name: ingress.Name, UID: ingress.UID},
```
sttts, line 517: what about existing objects with this owner ref?

deads2k:
> Do we need to convert objects using the old API, as [kubernetes/ingress-nginx@84102ee#diff-ece58b3bebec02a6d3173c8d8099d298R904-R939](https://github.com/kubernetes/ingress-nginx/commit/84102eec2ba270f624c57023aab59aab4471178e#diff-ece58b3bebec02a6d3173c8d8099d298R904-R939) does?

In repos that you own, yes you do.  `kubectl convert` may be able to do it for you.

pkg/route/ingress/ingress.go
deads2k: > what about existing objects with this owner ref?

They will still be fine until the old API is gone entirely, so you just have to scan for and update them.

ironcladlou:
@deads2k 
>>Do we need to convert objects using the old API, as kubernetes/ingress-nginx@84102ee#diff-ece58b3bebec02a6d3173c8d8099d298R904-R939 does?

>In repos that you own, yes you do. kubectl convert may be able to do it for you.

I'm not sure I asked my question very clearly. When the extensions API group finally "goes away", what exactly happens to the existing persisted resources in etcd? Must this controller (or some operator) be actively migrating the extensions API resources to the new API group to ensure that when the old API is gone forever there's nothing persisted in etcd with the old group?

ironcladlou:
@deads2k @sttts thoughts on https://github.com/openshift/openshift-controller-manager/pull/83#issuecomment-612913612?

sttts:
> I'm not sure I asked my question very clearly. When the extensions API group finally "goes away", what exactly happens to the existing persisted resources in etcd? Must this controller (or some operator) be actively migrating the extensions API resources to the new API group to ensure that when the old API is gone forever there's nothing persisted in etcd with the old group?

Nothing to do for operators or other consumers about the persisted api version. The persisted objects will be migrated through kube-storage-version-migrator (which we have an operator for now due to etcd encryption). This must be done before kube-apiserver stops being able to read the old api group. I trust @soltysh to remind us when this must be done.

ironcladlou:
Looks like this is ready to go then as far as I can tell. cc @openshift/sig-network-edge

pkg/route/ingress/ingress.go
```diff
@@ -8,8 +8,8 @@ import (
 
 	"k8s.io/klog"
 
-	"k8s.io/api/core/v1"
-	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	v1 "k8s.io/api/core/v1"
```
danehans, line 11: Either drop `v1` or rename to `corev1`.

ironcladlou: Fixed

danehans:
`imagestreams.image.openshift.io "pipeline" is forbidden: User "system:serviceaccount:ci:ci-operator" cannot get resource "imagestreams" in API group "image.openshift.io" in the namespace "ci-op-pb6fb9h4"`

CI flake?

/retest

PR #88, opened 2020-04-07
Bug 1819770: Bumping openshift/library-go

ricardomaraschini:


ricardomaraschini:
@gabemontero This is how it looks like when we bump library-go to release 4.4. We can bump it without bumping k8s.io. PTAL.

gabemontero:
> @gabemontero This is how it looks like when we bump library-go to release 4.4. We can bump it without bumping k8s.io. PTAL.

as we expected/hoped @ricardomaraschini :-) ... cool deal

dependencymagnet/doc.go
```diff
@@ -1,7 +1,9 @@
 package dependencymagnet
 
 import (
-	// alpha-build-machinery has non-go code needed for makefiles.
-	// Needed to work with go modules.
-	_ "github.com/openshift/library-go/alpha-build-machinery"
+	_ "github.com/openshift/build-machinery-go/make"
```
gabemontero, line 4: fyi I prefer this more precise, but multiple line, form but @adambkaplan called it out in my k8s rebase OCM PR, 
preferring just a single `"github.com/openshift/build-machinery-go"` entry

I'm happy with this as is but if you want to consolidate I won't object

Dockerfile.rhel
```diff
@@ -1,6 +1,5 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
 WORKDIR /go/src/github.com/openshift/openshift-controller-manager
-RUN yum install -y gpgme-devel libassuan-devel
```
gabemontero, line 3: folks looking at this fresh may be unaware, but these RPMs are already in the golang image we use as our builder/base
so this yum install was redundant at a minimum, and an opportunity for mirror flakes to mess up builds.

go.mod
```diff
@@ -1,46 +1,41 @@
...
-	k8s.io/api v0.0.0
-	k8s.io/apimachinery v0.0.0
-	k8s.io/apiserver v0.0.0
-	k8s.io/client-go v0.0.0
-	k8s.io/component-base v0.0.0
+	k8s.io/api v0.17.2
+	k8s.io/apimachinery v0.17.2
+	k8s.io/apiserver v0.17.1
+	k8s.io/client-go v0.17.1
+	k8s.io/component-base v0.17.1
```
gabemontero, line 38: these k8s entries match exactly what I see in https://github.com/openshift/library-go/blob/release-4.4/go.mod ... cool deal

ricardomaraschini:
@gabemontero I think I have seen @soltysh working on something very related to this(there is even a BZ), assigning him as well.

/assign @soltysh

dependencymagnet/doc.go
adambkaplan: as Gabe alluded, I prefer just a single import from `github.com/openshift/build-machinery-go`. This project is a bag of makefiles and scripts, so golang has no means of deciphering dependencies.

ricardomaraschini: Updated, PTAL.

Dockerfile
```diff
@@ -1,6 +1,5 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
```
soltysh, line 1: You'll need to update the build scripts in release too.

ricardomaraschini: You nailed it! https://github.com/openshift/release/pull/9029

PR #90, opened 2020-04-07
Report pod resource requests and limits as seen by scheduler and Kubelet

smarterclayton:
This is the first version of a feature we intend to add to Kubernetes as part of sig-instrumentation / sig-node / sig-scheduling collaboration.  OpenShift is an excellent test bed since we have had multiple complaints about this behavior.

A fundamental part of Kubernetes is resource consumption and requests as leveraged by the scheduler and enforced by the nodes. However, we have no metric that clearly represents the resource model and the calculations for those resources need to align to the standard Kube
model. In order to gain feedback from these metrics, we add the first version to OpenShift controller manager (which has all pods in memory anyway) and represent the metric as 'openshift_pod_resources'.

Administrators can use this metric in a number of ways to slice the actual usage of a cluster, and in the future we can use it to implement capacity planning graphs and alerts. Assuming the approach shown in OpenShift is adopted upstream with similar behavior we can adapt any consumers easily.

This metric is reasonably high cardinality, so only the active controller manager will expose them.

Initial discussion: https://github.com/kubernetes/kube-state-metrics/issues/1095
Upstream proposal draft: https://docs.google.com/document/d/13mjommm_wbwqKtisThmIa33TRa-gfIHJicxQ1S-TOlw/edit

At worst, we can carry this in 4.5 development branches and revert or fix before GA as that discussion continues.

### DIFF
--- a/pkg/cmd/controller/config.go
+++ b/pkg/cmd/controller/config.go
@@ -1,6 +1,8 @@
 package controller
 
 var ControllerInitializers = map[string]InitFunc{
+	"openshift.io/resource-metrics": RunResourceMetrics,
+
 	"openshift.io/serviceaccount": RunServiceAccountController,
 
 	"openshift.io/default-rolebindings": RunDefaultRoleBindingController,
--- a/pkg/cmd/controller/resources.go
+++ b/pkg/cmd/controller/resources.go
@@ -1,0 +1,10 @@
+package controller
+
+import (
+	"github.com/openshift/openshift-controller-manager/pkg/resources"
+)
+
+// RunResourceMetrics starts and registers the metrics for resource usage on a cluster.
+func RunResourceMetrics(ctx *ControllerContext) (bool, error) {
+	return true, resources.RegisterMetrics(ctx.KubernetesInformers.Core().V1().Pods().Lister())
+}
--- a/pkg/resources/resources.go
+++ b/pkg/resources/resources.go
@@ -1,0 +1,268 @@
+package resources
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/validation"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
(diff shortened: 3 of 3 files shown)


smarterclayton:
@brancz I think this is probably the best place for us to land until we reach an upstream decision.  OCM already watches all pods.

@adambkaplan @bparees this only losely couples to OCM, but is well isolated.  I think we understand the implications of the additional metrics, and if we find the better home in Kube we will move it there and then leave a recording rule in place to rename the metrics for anyone who is trapped.

smarterclayton:
A few examples from this PR:

Memory requested for openshift infrastructure per node minus usage (RSS) (positive numbers mean we are requesting more than we use)

```
sum by (node) (max without (instance,pod,namespace,job,endpoint,service) (openshift_pod_resources{lifecycle="",exported_namespace=~"openshift-.*",resource="memory",type="requests"})) - max by (node) (container_memory_rss{id="/"})
```

image

bparees:
only a 500meg gap between requested and used on most nodes, that seems like as close as we'd want to get?  (Though we should figure out what happened on that one node..guessing that means one bad component is requesting way more than it uses and landed on that node?)

smarterclayton:
yeah, basically this is showing the result of our 4.5 tuning which was more complicated because we lacked these fine grained yet accurate metrics

smarterclayton:
Any substantive objections to assessing the utility of this in master for 4.5?

bparees:
seems reasonable to me.

brancz:
I would like to actually compare this to the existing kube-state-metrics metrics, and see how different this would actually be. The metrics themselves look fine, but I'm not convinced yet that this is actually better than what we have and what we have is already deeply integrated into dashboards, alerting rules etc. and not to mention the rather high cardinality of metrics this adds to the setup.

smarterclayton:
Sure, can you detail someone to write up what those queries are and then i can review for equality given the underlying rules?  Init containers are not part of ksm today.  Also those are comments on the upstream proposal too.

adambkaplan:
@smarterclayton @brancz is this something we are going to hold until 4.6?

brancz:
We’re figuring out the way to move forward on upstream and can then follow up here.

adambkaplan:
@smarterclayton @brancz checking back in on this - should we close the PR?

brancz:
The conversation was started on upstream so I believe we can. We do need to make more progress on upstream but it was largely well received and understood.

PR #94, opened 2020-04-20
Bug 1815816: add 2 min retry on create build existing pod wrong owner (allow for GC)

gabemontero:
/assign @adambkaplan 

@bparees FYI in case you have cycles given your historical expertise

### DIFF
--- a/pkg/build/controller/build/build_controller.go
+++ b/pkg/build/controller/build/build_controller.go
@@ -1224,7 +1224,11 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
 		return update, fmt.Errorf("failed to create build pod: %v", err)
 
 	} else if err != nil {
-		bc.recorder.Eventf(build, corev1.EventTypeWarning, "FailedCreate", "Pod already exists: %s/%s", buildPod.Namespace, buildPod.Name)
+		// To facilitate usage scenarios often seen with automated scripting generating builds, where scripts
+		// 1. Create build
+		// 2. Pod is created
+		// 3. Delete build and immediately recreate it
+		// we want the build to wait for the old pod (from different build pod UID) to be cleaned up by GC. and then continues
 		klog.V(4).Infof("Build pod %s/%s for build %s already exists", build.Namespace, buildPod.Name, buildDesc(build))
 
 		// If the existing pod was not created by this build, switch to the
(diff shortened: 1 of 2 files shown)


bparees:
would it be simpler to just always retry creating the new build pod for 2 minutes before giving up?

the case where an old pod exists and should not and will not be cleaned up automatically should be exceedingly rare, i'm ok w/ someone having to wait 2 minutes to be told their build can't proceed under those circumstances.

gabemontero:
> would it be simpler to just always retry creating the new build pod for 2 minutes before giving up?
> 
> the case where an old pod exists and should not and will not be cleaned up automatically should be exceedingly rare, i'm ok w/ someone having to wait 2 minutes to be told their build can't proceed under those circumstances.

Are you saying we should remove the existing logic that tries to use the pod if the owner ref matches?

If not, sorry, I'll need some more clarification.  Thanks.

bparees:
> Are you saying we should remove the existing logic that tries to use the pod if the owner ref matches?

no, just saying if a pod exists that does not have the right ownerref, let's go into a retry for 2 mins, instead of immediately aborting.  We can just retry based on the age (Creation time) of the build resource, rather than having to muck w/ adding new annotations to resources to track timestamps.

unless i'm not fully understanding the failure scenario here?

gabemontero:
> > Are you saying we should remove the existing logic that tries to use the pod if the owner ref matches?
> 
> no, just saying if a pod exists that does not have the right ownerref, let's go into a retry for 2 mins, instead of immediately aborting. We can just retry based on the age (Creation time) of the build resource, rather than having to muck w/ adding new annotations to resources to track timestamps.
> 
> unless i'm not fully understanding the failure scenario here?

ok working off of the build creation time does seem sufficient and much less complicated than the separate annotation

I'll re-work based on that and we can go from there.

thanks

gabemontero:
update pushed @bparees - annotation removed and creation timestamp used

bparees:
/approve

it looks reasonable to me, but will let @adambkaplan give it a final review.

pkg/build/controller/build/build_controller.go
```diff
@@ -1224,7 +1224,11 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
 		return update, fmt.Errorf("failed to create build pod: %v", err)
 
 	} else if err != nil {
-		bc.recorder.Eventf(build, corev1.EventTypeWarning, "FailedCreate", "Pod already exists: %s/%s", buildPod.Namespace, buildPod.Name)
+		// To facilitate usage scenarios often seen with automated scripting generating builds, where scripts
+		//1. Create build
+		//2. Pod is created
+		//3. Delete build and immediately recreate it
+		//we want the build to wait for the old pod (from different build pod UID) to be cleaned up by GC. and then continues
```
adambkaplan, line 1231: nit - missing spaces after //

pkg/build/controller/build/build_controller.go
```diff
@@ -1233,11 +1237,30 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
...
 			return nil, err
 		}
+		// HashOwnerReference does not distinguish between having no refs to compare with the build and
+		// having a different ref compared to our build here;  that said, if the pod has not owner refs,
+		// there is no chance GC will clean it up, so bypass our retry here
 		if !strategy.HasOwnerReference(existingPod, build) {
+			if len(existingPod.OwnerReferences) > 0 {
+				// check to see if we have retried long enough for the old pod to get GC'ed
+				now := metav1.Now()
+				twoMinutesAgo := now.Add(-2 * time.Minute)
```
adambkaplan, line 1247: nit - can simplify to 
```suggestion
				twoMinutesAgo := metav1.Now().Add(-2 * time.Minute)
```
no?

gabemontero:
nits addressed @adambkaplan thanks PTAL

PR #96, opened 2020-04-27
Bug 1815816: more timing windows on build/pod owner ref correlation (allow GC to c…

gabemontero:
…lean up old pods)

QE testing and subsequent retries by me found a timing window we missed with https://github.com/openshift/openshift-controller-manager/pull/94

/assign @adambkaplan 

@bparees fyi


pkg/build/controller/build/build_controller.go
```diff
@@ -1087,6 +1092,21 @@ func (bc *BuildController) resolveImageReferences(build *buildv1.Build, update *
 	return nil
 }
 
+func retryOnOwnerRef(build *buildv1.Build, pod *corev1.Pod) error {
+	if len(pod.OwnerReferences) > 0 {
```
soltysh, line 1096: A general rule is to return early, the code will read better if this looks like:
```
if len(pod.OwnerReferences) == 0 {
    return nil
}
// ... rest of the code
```

PR #99, opened 2020-05-04
Bug 1831243: v1 StatefulSets do not trigger on image changes

smarterclayton:
We must check for v1 StatefulSets during update.


smarterclayton:
4.4 bug is 1831245 and 4.3 bug is 1831246

PR #105, opened 2020-05-11
Bug 1829447: ensure build pod activeDeadlineSeconds always set

gabemontero:
RunOnceDuration admission plugin no longer present in 4.x

/assign @adambkaplan

### DIFF
--- a/pkg/build/controller/strategy/util.go
+++ b/pkg/build/controller/strategy/util.go
@@ -289,6 +289,23 @@ func addOutputEnvVars(buildOutput *corev1.ObjectReference, output *[]corev1.EnvV
 	return nil
 }
 
+// setupActiveDeadline sets up the Pod activeDeadlineSeconds field
+func setupActiveDeadline(pod *corev1.Pod, build *buildv1.Build) *corev1.Pod {
+	if build.Spec.CompletionDeadlineSeconds != nil {
+		pod.Spec.ActiveDeadlineSeconds = build.Spec.CompletionDeadlineSeconds
+		return pod
+	}
+
+	// RunOnceDuration admission plugin was used to include the default active deadline for run-once pods, like the build pods
+	// but it was removed from OpenShift in 4.0; rather than ship the RunOnceDuration admission as webhook admission
+	// plugin, which will involve creating new operator, we use a long activeDeadlineSeconds as build are
+	// designed to terminate
+	var defActiveDeadline int64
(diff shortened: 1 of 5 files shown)


gabemontero:
all green tests @adambkaplan

pkg/build/controller/strategy/util.go
```diff
@@ -289,6 +289,22 @@ func addOutputEnvVars(buildOutput *corev1.ObjectReference, output *[]corev1.EnvV
...
+		pod.Spec.ActiveDeadlineSeconds = build.Spec.CompletionDeadlineSeconds
+	} else {
+		// RunOnceDuration admission plugin was used to include the default active deadline for run-once pods, like the build pods
+		// but it was removed from OpenShift in 4.0; rather than ship the RunOnceDuration admission as webhook admission
+		// plugin, which will involve creating new operator, we use a long activeDeadlineSeconds as build are
+		// designed to terminate
+		var defActiveDeadline int64
+		defActiveDeadline = 604800 // 1 week = 60 sec * 60 min * 24 hr * 7 days
+		pod.Spec.ActiveDeadlineSeconds = &defActiveDeadline
+	}
```
adambkaplan, line 304: golang style nit. +1 for returning the object you mutate, though!

```suggestion
	if build.Spec.CompletionDeadlineSeconds != nil {
		pod.Spec.ActiveDeadlineSeconds = build.Spec.CompletionDeadlineSeconds
		return pod
	}
	// RunOnceDuration admission plugin was used to include the default active deadline for run-once pods, like the build pods
	// but it was removed from OpenShift in 4.0; rather than ship the RunOnceDuration admission as webhook admission
	// plugin, which will involve creating new operator, we use a long activeDeadlineSeconds as build are
	// designed to terminate
	var defActiveDeadline int64
	defActiveDeadline = 604800 // 1 week = 60 sec * 60 min * 24 hr * 7 days
	pod.Spec.ActiveDeadlineSeconds = &defActiveDeadline
```

gabemontero:
style nit incorporated @adambkaplan ptal

gabemontero:
all green tests again @adambkaplan

gabemontero:
> Minor style nit.
> 
> As a follow up (beyond the scope of this PR and BZ) - I noticed that we don't have `CompletionDeadlineSeconds` as an option for our build defaults. If a customer wants to tune the active deadline for the whole cluster, it may be worth adding this as an option.

good point @adambkaplan .... though with where we are with builds in general my initial reaction is to wait for the ask from a customer at this point .... my 2 cents at least

PR #107, opened 2020-05-13
Bug 1809719: use ReleaseOnCancel for leader election config

gabemontero:
/assign @adambkaplan 

@smarterclayton fyi

### DIFF
--- a/pkg/cmd/openshift-controller-manager/controller_manager.go
+++ b/pkg/cmd/openshift-controller-manager/controller_manager.go
@@ -94,10 +94,11 @@ func RunOpenShiftControllerManager(config *openshiftcontrolplanev1.OpenShiftCont
 	}
 	go leaderelection.RunOrDie(context.Background(),
 		leaderelection.LeaderElectionConfig{
-			Lock:          rl,
-			LeaseDuration: config.LeaderElection.LeaseDuration.Duration,
-			RenewDeadline: config.LeaderElection.RenewDeadline.Duration,
-			RetryPeriod:   config.LeaderElection.RetryPeriod.Duration,
+			Lock:            rl,
+			ReleaseOnCancel: true,
+			LeaseDuration:   config.LeaderElection.LeaseDuration.Duration,
+			RenewDeadline:   config.LeaderElection.RenewDeadline.Duration,
+			RetryPeriod:     config.LeaderElection.RetryPeriod.Duration,
 			Callbacks: leaderelection.LeaderCallbacks{
 				OnStartedLeading: originControllerManager,
 				OnStoppedLeading: func() {


gabemontero:
all green tests @adambkaplan

adambkaplan:
@gabemontero can you please demonstrate that OCM gracefully terminates? I feel like this would be hard to instrument in CI, but could be proved manually:

0. List the OCM pods
1. Find the leader OCM pod
2. Delete the leader OCM pod
3. Verify that one of the remaining OCM pods obtains leader election in a short period of time (say < 10 seconds?)

gabemontero:
> @gabemontero can you please demonstrate that OCM gracefully terminates? I feel like this would be hard to instrument in CI, but could be proved manually:
> 
>     1. List the OCM pods
> 
>     2. Find the leader OCM pod
> 
>     3. Delete the leader OCM pod
> 
>     4. Verify that one of the remaining OCM pods obtains leader election in a short period of time (say < 10 seconds?)

will do @adambkaplan ..... headed over to cluster-bot

gabemontero:
Bottom line, I think we are good @adambkaplan 

However, there is a possibly tl;dr explanation on how I get to that determination.

1) A delete alone is taking 40 seconds.  If you start after the delete, it takes a minute for the new leader to get the lease.

```
gmontero ~ $ date ; oc delete pod controller-manager-47d5c ; date 
Thu 14 May 2020 12:12:28 PM EDT
pod "controller-manager-47d5c" deleted
Thu 14 May 2020 12:13:09 PM EDT
gmontero ~ $ 
```

and account for 6 hr TZ diff between EDT and UTC

```
oc logs -f controller-manager-96z6p  # existing pod
...
I0514 15:59:30.904899       1 leaderelection.go:242] attempting to acquire leader lease  openshift-controller-manager/openshift-master-controllers...
I0514 16:14:10.488034       1 leaderelection.go:252] successfully acquired lease openshift-controller-manager/openshift-master-controllers
```

2) But then there is the other LeaderElectionConfig to consider ... see https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L111-L157

The leader election config for OCM is in the configmap `openshift-master-controllers`.  If you look at the contents of it in my test cluster, it is:

```
gmontero ~ $ oc get cm openshift-master-controllers -o yaml 
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    control-plane.alpha.kubernetes.io/leader: '{"holderIdentity":"controller-manager-96z6p","leaseDurationSeconds":60,"acquireTime":"2020-05-14T16:14:10Z","renewTime":"2020-05-14T16:40:32Z","leaderTransitions":4}'
  creationTimestamp: "2020-05-14T15:42:53Z"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:metadata:
        f:annotations:
          .: {}
          f:control-plane.alpha.kubernetes.io/leader: {}
    manager: openshift-controller-manager
    operation: Update
    time: "2020-05-14T16:40:32Z"
  name: openshift-master-controllers
  namespace: openshift-controller-manager
  resourceVersion: "39822"
  selfLink: /api/v1/namespaces/openshift-controller-manager/configmaps/openshift-master-controllers
  uid: cc23d4e2-92b6-4996-a7d2-349fffc79e94
gmontero ~ $ 
```

3) the `leaseDurationSeconds:60` is the key here.  Per the godoc from my LeaderElectionConfig link above, that is how long non-leader candidates will wait to force leadership:

```
	// LeaseDuration is the duration that non-leader candidates will
	// wait to force acquire leadership. This is measured against time of
	// last observed ack.
```

At 60 seconds then, the 1 minute gap I saw means it happened pretty instantaneously.  Certainly less than 10 seconds.

In fact, if you consider this from the godoc:

```
	// RetryPeriod is the duration the LeaderElector clients should wait
	// between tries of actions.
	//
	// Core clients default this value to 2 seconds.
	RetryPeriod time.Duration
```

and the fact that we are not setting the retry period, you would expect 62 seconds between the leader dying and the backup taking over.  My run above was in 61 seconds.

adambkaplan:
/cc @smarterclayton @deads2k 

I'm not entirely sold that this change is fixing the bug to our conclusion. 40 seconds feels like a long time for the controller to shut down - can we tell if SIGKILL was sent in addition to SIGTERM? The former implies non-graceful termination.

gabemontero:
In parallel, I'll do some compares of the time to delete, time for new leader to acquire, on a cluster without this change.  I'll also see about getting some thread dumps, though I've never attempted that while deleting a pod and so am not 100% confident I'll get anything useful.

gabemontero:
so the timings on how long 1) the delete of the leader pod took, and b) how long it took someone else to become leader were identical with a 4.5 build without this change

I'm still unclear on whether those timings have any bearing on whether the change requested is doing what is expected 

I could not get kill SIGQUIT to generate a thread dump.  Either that is the wrong signal, though that is what my internet searches for golang point to, or I need to add code to OCM, a la https://stackoverflow.com/questions/19094099/how-to-dump-goroutine-stacktraces

I'll give that a go when I next circle back to this, unless course correction is provided before I do.

@adambkaplan @smarterclayton @deads2k

gabemontero:
OK I was able to patch the OCM and trigger a handful of thread dumps via `kill -3 1` while deleting the leader pod @adambkaplan @smarterclayton @deads2k 

I've attached here for any interest.  I'll start analyzing this afternoon.

[tdumps.txt](https://github.com/openshift/openshift-controller-manager/files/4646199/tdumps.txt)

gabemontero:
OK the thread dump was helpful for me @adambkaplan 

My bottom line: we should merge this and move on.

tl;dr

1) from a leader election perspective, all three dumps show it sitting here in goroutine 101:  https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L262-L279

2) meaning that thread will keep the lock until the process shutdown cancels that thread's context

3) given all the other background threads in the OCM for the various controllers that you see in the dump, in hindsight I am not surprised it takes some time to clean all that up

4) the new bit we sit then triggers here in the same method once that wait is interrupted, namely here:  https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L281-L284

5) which means we release the lock before shutting down here:  https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L288-L302

6) which removing *possible* contention at this path here:  https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L332-L344

7) based on the various intervals, the time spent churning over https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/leaderelection.go#L332-L344 may not be super perceptable with our prior experiment

8) nor will this setting in anyway speed up shutdown, per my points 2), 3), and 4)

adambkaplan:
/lgtm

based on the analysis by @gabemontero we are doing the best we can here.

PR #111, opened 2020-05-21
Bug 1832294: Add strategy metrics

tnozicka:
Add `strategy_total` metric to identify usage for different strategies. Also fixed a bug in counting `last_failed_rollout_time`.

Requires:
- [x] unit test

/cc @soltysh

### DIFF
--- a/pkg/apps/metrics/prometheus/metrics.go
+++ b/pkg/apps/metrics/prometheus/metrics.go
@@ -2,11 +2,14 @@ package prometheus
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/blang/semver"
+	appsv1 "github.com/openshift/api/apps/v1"
+	appsv1listers "github.com/openshift/client-go/apps/listers/apps/v1"
 	"github.com/prometheus/client_golang/prometheus"
 
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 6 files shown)


pkg/apps/metrics/prometheus/metrics.go
```diff
@@ -145,12 +157,13 @@ func (c *appsCollector) Collect(ch chan<- prometheus.Metric) {
 				continue
 			}
 			if util.IsCompleteDeployment(d) {
-				// If a completed rollout is found AFTER we recorded a failed rollout,
-				// do not record the lastFailedRollout as the latest rollout is not
-				// failed.
-				if r, hasFailedRollout := latestFailedRollouts[key]; hasFailedRollout && r.latestVersion < latestVersion {
-					delete(latestFailedRollouts, key)
```
tnozicka, line 152: this was broken as it was dependent on the order of seeing the objects listed

pkg/apps/metrics/prometheus/metrics_test.go
```diff
@@ -359,24 +443,43 @@ func TestLegacyRegistry(t *testing.T) {
...
-	ac := appsCollector{
-		lister: lister,
-		nowFn:  defaultTimeNowFn,
+	registry := metrics.NewKubeRegistry()
+	err := registry.Register(&collector)
+	if err != nil {
+		t.Fatal(err)
 	}
 
-	legacyregistry.MustRegister(&ac)
```
tnozicka, line 369: had to change this because it was panicking if the test was run more then once

PR #112, opened 2020-05-22
Bug 1829786: Allow scheduled retry of failed ImageStream import

ricardomaraschini:
If an ImageStream tag refers to an image by SHA and we fail to import it
we need to allow the schedule process to take over an try again. With this
patch we now retry to import image stream tags based on SHA until it succeeds
at least once.

For further info on the reasoning behind this PR, please check https://coreos.slack.com/archives/CE2HALN2W/p1590050918461900


ricardomaraschini:
/retitle Bug 1829786: Allow scheduled retry of failed ImageStream import

PR #115, opened 2020-05-29
Reflect route status to ingress

miheer:
When the status of a route corresponding to an ingress is updated, the ingress's status should be updated as well.

https://issues.redhat.com/browse/NE-27

### DIFF
--- a/pkg/cmd/controller/ingress.go
+++ b/pkg/cmd/controller/ingress.go
@@ -5,6 +5,9 @@ import (
 
 	routeclient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/openshift/openshift-controller-manager/pkg/route/ingress"
+
+	ingressclient "k8s.io/client-go/kubernetes"
+
 )
 
 func RunIngressToRouteController(ctx *ControllerContext) (bool, error) {
@@ -17,10 +20,15 @@ func RunIngressToRouteController(ctx *ControllerContext) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+	ingressClient, err := ingressclient.NewForConfig(clientConfig)
+	if err != nil {
+		return false, err
+	}
 
 	controller := ingress.NewController(
 		coreClient,
 		routeClient,
+		ingressClient,
 		ctx.KubernetesInformers.Networking().V1beta1().Ingresses(),
 		ctx.KubernetesInformers.Core().V1().Secrets(),
 		ctx.KubernetesInformers.Core().V1().Services(),
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	networkingv1beta1informers "k8s.io/client-go/informers/networking/v1beta1"
+	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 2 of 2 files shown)


miheer:
@Miciah PTAL

pkg/cmd/controller/ingress.go
```diff
@@ -5,6 +5,9 @@ import (
 
 	routeclient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	"github.com/openshift/openshift-controller-manager/pkg/route/ingress"
+
+	ingressclient "k8s.io/client-go/kubernetes"
+
```
Miciah, line 10: The blank line (line 10) can be deleted.

pkg/route/ingress/ingress.go
```diff
@@ -405,6 +409,24 @@ func (c *Controller) sync(key queueKey) error {
...
 			if routeMatchesIngress(existing, ingress, &rule, &path, c.secretLister, c.serviceLister) {
+				var loadbalancerIngresses []corev1.LoadBalancerIngress
+				for _, rule := range ingress.Spec.Rules {
+
+					loadbalancerIngress := corev1.LoadBalancerIngress{
+						Hostname: rule.Host,
+					}
+
+					loadbalancerIngresses = append(loadbalancerIngresses, loadbalancerIngress)
+
```
Miciah, line 420: Some of these blank lines can be deleted (lines 414, 418, and 420).

pkg/route/ingress/ingress.go
```diff
@@ -405,6 +409,24 @@ func (c *Controller) sync(key queueKey) error {
 			}
 
 			if routeMatchesIngress(existing, ingress, &rule, &path, c.secretLister, c.serviceLister) {
+				var loadbalancerIngresses []corev1.LoadBalancerIngress
+				for _, rule := range ingress.Spec.Rules {
```
Miciah, line 413: Shouldn't this be using the status of the route?  I would expect the logic to use `old` or `existing` and collect `old[i].Status.Ingress[j].RouterCanonicalHostname` for the ingress status.  If we take the approach that I am suggesting, the logic should also uniquify the list, or use `sets.String` when collecting names, because one ingress may have many routes that are all exposed by the same router.

Would it make more sense to do this before the loop, on line 387?  We already have the list of routes that correspond to the ingress (as indicated by their owner references) there; putting the new logic here actually seems to complicate things.

The semantics gets iffy if, for example, an ingress is translated into multiple routes, one of those routes is admitted by some but not all routers, and another route is admitted by some router that did not admit the first route.  However, I think collecting all canonical names of all routers that expose at least one of the ingress's routes is the most reasonable translation within the constraints of the API.

pkg/route/ingress/ingress.go
```diff
@@ -405,6 +409,24 @@ func (c *Controller) sync(key queueKey) error {
...
+					}
+
+					loadbalancerIngresses = append(loadbalancerIngresses, loadbalancerIngress)
+
+				}
+
+				ingress.Status.LoadBalancer = corev1.LoadBalancerStatus{Ingress: loadbalancerIngresses}
+
+				ingClient := c.ingressClient.NetworkingV1beta1().Ingresses(ingress.Namespace)
+				_, err = ingClient.UpdateStatus(context.TODO(), ingress, metav1.UpdateOptions{})
```
Miciah, line 426: The logic should check whether an update is needed.

Miciah:
Superseded by https://github.com/openshift/openshift-controller-manager/pull/130.
/close

PR #117, opened 2020-06-10
Bug 1832839: more changes required for named workqueue metrics on pre 1.18 k8s base

gabemontero:
attempting to implement guidance from https://bugzilla.redhat.com/show_bug.cgi?id=1832839#c10

### DIFF
--- a/pkg/metrics/workqueue.go
+++ b/pkg/metrics/workqueue.go
@@ -1,0 +1,130 @@
+package metrics
+
+import (
+	"k8s.io/client-go/util/workqueue"
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// Per recommendation from the openshift monitoring team, this module has been
+// copied from https://github.com/openshift/openshift-controller-manager/blob/master/vendor/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go
+// and then adjusted for use within OCM for adding metrics for the named OCM workqueues; as you can tell from the URL,
+// the module exists in upstream k8s, but only at versions 1.18 and greater, so
+// we are approximating the function / intent for versions of openshift based on earlier
+// versions of k8s, where we still want those metrics for named workqueues
+
+// Metrics subsystem and keys used by the workqueue.
+const (
+	WorkQueueSubsystem         = "workqueue"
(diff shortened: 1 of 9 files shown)


gabemontero:
all green tests @adambkaplan

pkg/metrics/workqueue.go
```diff
@@ -0,0 +1,126 @@
+package metrics
+
+import (
+	"k8s.io/client-go/util/workqueue"
+	k8smetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// Package prometheus sets the workqueue DefaultMetricsFactory to produce
+// prometheus metrics. To use this package, you just have to import it.
```
adambkaplan, line 10: We need to doc that this is a port of upstream code that landed in a future version. I assume there was a lot of copy/pasting here?

gabemontero: comment updated @adambkaplan

gabemontero:
all green tests @adambkaplan 

> Nit on the doc - we also need some kind of automated test to ensure we don't regress.

Can you elaborate a bit more on the automated test you are thinking about here @adambkaplan ?

thanks

adambkaplan:
@gabemontero unfortunately the best item I could come up with is to add e2e's to origin. We have existing tests for build metrics in the `prometheus` subdirectory.

Example:
https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus_builds.go

gabemontero:
> @gabemontero unfortunately the best item I could come up with is to add e2e's to origin. We have existing tests for build metrics in the `prometheus` subdirectory.
> 
> Example:
> https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus_builds.go

Ah OK @adambkaplan thanks for clarifying what you meant.

And I know the guy who originally created those tests back in 3.10, so I can talk to him if need be ..... ebaG oretnoM or somebody like that :-)

gabemontero:
So the 4.4 e2e added in https://github.com/openshift/origin/pull/25117 fails as expected @adambkaplan with this PR not merging.

Stats like those noted in https://github.com/openshift/openshift-controller-manager/pull/117#issuecomment-642212971 come up as unsuccessful since the metrics are missing.

PR #123, opened 2020-06-25
Jira build 96 add forcepull to buildoverrides

coreydaley:
Use the ForcePull pointer from BuildOverridesConfig

https://issues.redhat.com/browse/BUILD-96


pkg/build/controller/build/overrides/overrides.go
```diff
@@ -32,22 +32,28 @@ func (b BuildOverrides) ApplyOverrides(pod *corev1.Pod) error {
...
 		}
 		if build.Spec.Strategy.SourceStrategy != nil {
-			klog.V(5).Infof("Setting source strategy ForcePull to true in build %s/%s", build.Namespace, build.Name)
-			build.Spec.Strategy.SourceStrategy.ForcePull = true
+			klog.V(5).Infof("Setting source strategy ForcePull to %t in build %s/%s", *b.Config.ForcePull, build.Namespace, build.Name)
+			build.Spec.Strategy.SourceStrategy.ForcePull = *b.Config.ForcePull
 		}
 		if build.Spec.Strategy.CustomStrategy != nil {
-			err := applyForcePullToPod(pod)
+			pullPolicy := corev1.PullNever
```
adambkaplan, line 45: This should be `corev1.IfNotPresent`.

coreydaley: Updated, thanks.

pkg/build/controller/build/overrides/overrides.go
```diff
@@ -32,22 +32,28 @@ func (b BuildOverrides) ApplyOverrides(pod *corev1.Pod) error {
 
 	klog.V(4).Infof("Applying overrides to build %s/%s", build.Namespace, build.Name)
 
-	if b.Config.ForcePull {
+	if b.Config.ForcePull != nil {
```
gabemontero, line 35: could we change this to `if b.Config.ForcePull != nil && *b.Config.ForcePull` and then leave most of what is below the same?

pkg/build/controller/build/overrides/overrides.go
gabemontero: wasn't this resolved incorrectly ... the code change still shows defaulting to Never 

Per https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2243 the default is IfNotPresent anyways

seems like you should leave `applyForcePullToPod` as is and do not add the pullPolicy param

gabemontero:
in case it is masked by github's PR review ... I think the comment from 13 days ago around line 45 of overrides.go was resolved incorrectly @coreydaley @adambkaplan 

the default should be IfNotPresent, which is what k8s go doc says, and hence I don't think `applyForcePullToPod` should need modification

coreydaley:
If the system admin has set ForcePull to false, shouldn't we propagate that down also?
So if a user has set it to corev1.PullAlways, and the sys admin set ForcePull to false, shouldn't that change the users config to corev1.IfNotPresent?

pkg/build/controller/build/overrides/overrides.go
gabemontero: duh, nevermind ... if they switch back to false ... see @coreydaley 's https://github.com/openshift/openshift-controller-manager/pull/123#issuecomment-655558903

feel free to resolve this one @coreydaley

gabemontero:
> If the system admin has set ForcePull to false, shouldn't we propagate that down also?

duh ... correct ... also updated my comment above ... feel free to resolve

> So if a user has set it to corev1.PullAlways, and the sys admin set ForcePull to false, shouldn't that change the users config to corev1.IfNotPresent?

ok for this one, the user does not have access to set pull always, etc. ... they just control force pull, and force pull should either translate to always or ifnotpresent 

Your code still has `pullPolicy := corev1.PullNever`

coreydaley:
> Your code still has `pullPolicy := corev1.PullNever`

Yep, I will fix that, I think it got reset when I made a recent change :(

coreydaley:
@gabemontero changed to IfNotPresent instead of Never

pkg/build/controller/build/overrides/overrides_test.go
```diff
@@ -34,7 +34,8 @@ func TestBuildOverrideForcePull(t *testing.T) {
 	ops := []admission.Operation{admission.Create, admission.Update}
 	for _, test := range tests {
 		for _, op := range ops {
-			overrides := BuildOverrides{Config: &openshiftcontrolplanev1.BuildOverridesConfig{ForcePull: true}}
+			truePtr := true
```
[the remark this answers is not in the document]
coreydaley: updated and pushed

gabemontero:
LGTM.   Let’s wait and see if @adambkaplan wants to chime in.  If he
doesn’t by scrum tomorrow we’ll ask him then.

coreydaley:
@adambkaplan Do you have any comments on this?

pkg/build/controller/build/overrides/overrides.go
```diff
@@ -98,14 +104,14 @@ func (b BuildOverrides) ApplyOverrides(pod *corev1.Pod) error {
 	return common.SetBuildInPod(pod, build)
 }
 
-func applyForcePullToPod(pod *corev1.Pod) error {
+func applyForcePullToPod(pod *corev1.Pod, pullPolicy corev1.PullPolicy) error {
```
adambkaplan, line 107: nit - can rename this to `applyPullPolicyToPod`

coreydaley: updated in current push

coreydaley:
@adambkaplan @gabemontero all comments have been addressed and all requested updates completed. ptal

PR #124, opened 2020-07-09
Route ingress: Allow setting termination policy via annotation

alvaroaleman:
Ingress objects are more common in the Kube ecosystem and have deeper
integration with other components. Right now they can not be used when a
terminationPolicy other than Edge is required. This PR changes that by
allowing to set an annotation on the Ingress.

cc @hongkailiu

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -448,10 +448,7 @@ func (c *Controller) sync(key queueKey) error {
 		}
 	}
 
-	if len(errs) > 0 {
-		return utilerrors.NewAggregate(errs)
-	}
-	return nil
+	return utilerrors.NewAggregate(errs)
 }
 
 func hasIngressOwnerRef(owners []metav1.OwnerReference) (string, bool) {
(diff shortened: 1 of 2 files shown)


Miciah:
This looks good, but I'd like to discuss it with the team before merging it.

By the way, during review, I was confused about `insecureEdgeTerminationPolicy`.  Although the API documentation implies that `insecureEdgeTerminationPolicy` is applicable only to edge-terminated routes (https://github.com/openshift/api/blob/af4dd20aed231f81f1c76f0a98dc60947f12a7a4/route/v1/types.go#L250-L257), the validation allows `insecureEdgeTerminationPolicy` to be used with edge, reencrypt, and passthrough, with the caveat that `insecureEdgeTerminationPolicy: Allow` is prohibited for passthrough routes (https://github.com/openshift/openshift-apiserver/blob/3dc75d3dd5b8de9487122d4a798ab1600217efb8/pkg/route/apis/route/validation/validation.go#L243-L264).

alvaroaleman:
@Miciah yeah that we can not validate and provide feedback is not ideal. Does openshift already have a webhook or anything else that can be used to add additional validations for objects served by the kube-apiserver, do you know?

Miciah:
> Does openshift already have a webhook or anything else that can be used to add additional validations for objects served by the kube-apiserver, do you know?

I'm not aware of anything, sorry.  I'd ask someone from the API team.

hongkailiu:
@Miciah 

Is it possible to backport it 4.5?

alvaroaleman:
@hongkailiu right now there is a follow-up open to fix some issues with this: https://github.com/openshift/openshift-controller-manager/pull/128 That would need to merge first

PR #126, opened 2020-07-16
Node Image Storage (PoC)

otaviof:
This pull-request implements the changes described by [Node Cache for Builds enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/builds/node-cache-for-builds.md), bringing read-only access to node images for builds. Please consider the document for full details.

The modifications were mostly in `setupContainersNodeStorage` and `setupContainersStorage` in order to mount volumes as described bellow.

## Volumes and Mounts

In the following table you can see how the volumes are mounted on build pods.

| Name                   | Node (`hostPath`)             | Pod                           | Mode |
|------------------------|-------------------------------|-------------------------------|------|
| node-storage-root      | `/var/lib/containers/storage` | `/var/lib/shared`             | `ro` |
| container-storage-root | None (`EmptyDir{}`)           | `/var/lib/containers/storage` | `rw` |

The mount location `/var/lib/shared` is chosen due the existing [configuration in openshift/builder](https://github.com/openshift/builder/blob/00f9e90d06c3ac8aeac9642e2af3221eb7f0b6e1/imagecontent/etc/containers/storage.conf#L22). This location could be changed later on, however, that's also the default location for `buildah` upstream.

## OpenShift-Builder

With [current `containers/storage`](https://github.com/openshift/builder/blob/00f9e90d06c3ac8aeac9642e2af3221eb7f0b6e1/go.mod#L10), it will end-up with permission issues on trying to access node layers. Like the following example:

```
error: build error: Failed to push image: error copying layers and metadata from "containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.imagestore=/var/lib/shared]image-registry.openshift-image-registry.svc:5000/otaviof/rails-ex:latest" to "docker://image-registry.openshift-image-registry.svc:5000/otaviof/rails-ex:latest": Error reading blob sha256:2bf094d88b12300b2226c4c24f80533dec869141b52b88f706843329c4f54b1d: error reading blob from source image "containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.imagestore=/var/lib/shared]image-registry.openshift-image-registry.svc:5000/otaviof/rails-ex:latest": not allowed to update mount locations for layers at "/var/run/containers/storage/overlay-layers/mountpoints.json": called a write method on a read-only store
```

Therefore, this pull-request is linked with [PR #164 in openshift-builder](https://github.com/openshift/builder/pull/164).

## Testing

With both images for openshift-controller-manager and openshift-builder (based on [PR #164](https://github.com/openshift/builder/pull/164)), create a new `BuildConfig` object with:

```bash
oc new-app --name=nodejs-ex https://github.com/sclorg/nodejs-ex.git
```

Then, add a distinct label in one of the cluster nodes. For instance:

```bash
oc label nodes ip-10-0-139-172.ec2.internal test-build=true
```

And edit `BuilldConfig` to use the same label:

```bash
oc patch buildconfig/nodejs-ex --patch='{ "spec": { "nodeSelector": { "test-build": "true" } } }'
```

Finally, the `nodejs-ex` build can be started:

```bash
oc start-build nodejs-ex
```

### DIFF
--- a/pkg/build/controller/strategy/util.go
+++ b/pkg/build/controller/strategy/util.go
@@ -463,7 +463,7 @@ func setupContainersStorage(pod *corev1.Pod, container *corev1.Container) {
 }
 
 // setupContainersNodeStorage borrows the appropriate storage directories from the node so
-// that we can share layers that we're using with the node, as a additional read-only
+// that we can share layers that we're using with the node, as an additional read-only
 // mount point.
 func setupContainersNodeStorage(pod *corev1.Pod, container *corev1.Container) {
 	exists := false
--- a/pkg/build/controller/strategy/util.go
+++ b/pkg/build/controller/strategy/util.go
@@ -43,6 +43,11 @@ const (
 	// GitCloneContainer is the name of the container that will clone the
 	// build source repository and also handle binary input content.
 	GitCloneContainer = "git-clone"
+
+	// containersStoragePath is the full path to default images storage location.
+	containersStoragePath = "/var/lib/containers/storage"
+	// containersAdditionalStoragePath is the full path to additional images storage location.
+	containersAdditionalStoragePath = "/var/lib/shared"
 )
 
 const (
@@ -450,7 +455,7 @@ func setupContainersStorage(pod *corev1.Pod, container *corev1.Container) {
 	container.VolumeMounts = append(container.VolumeMounts,
 		corev1.VolumeMount{
 			Name:      "container-storage-root",
-			MountPath: "/var/lib/containers/storage",
+			MountPath: containersStoragePath,
 		},
 	)
 	container.Env = append(container.Env, corev1.EnvVar{Name: "BUILD_STORAGE_DRIVER", Value: "overlay"})
@@ -458,7 +463,8 @@ func setupContainersStorage(pod *corev1.Pod, container *corev1.Container) {
 }
 
 // setupContainersNodeStorage borrows the appropriate storage directories from the node so
-// that we can share layers that we're using with the node
+// that we can share layers that we're using with the node, as a additional read-only
+// mount point.
 func setupContainersNodeStorage(pod *corev1.Pod, container *corev1.Container) {
 	exists := false
 	for _, v := range pod.Spec.Volumes {
@@ -468,23 +474,26 @@ func setupContainersNodeStorage(pod *corev1.Pod, container *corev1.Container) {
 		}
 	}
 	if !exists {
+		directoryOrCreate := corev1.HostPathDirectoryOrCreate
 		pod.Spec.Volumes = append(pod.Spec.Volumes,
-			// TODO: run unprivileged https://github.com/openshift/origin/issues/662
 			corev1.Volume{
 				Name: "node-storage-root",
 				VolumeSource: corev1.VolumeSource{
 					HostPath: &corev1.HostPathVolumeSource{
-						Path: "/var/lib/containers/storage",
+						Type: &directoryOrCreate,
+						// default storage path for node images
+						Path: containersStoragePath,
 					},
 				},
 			},
 		)
 	}
 	container.VolumeMounts = append(container.VolumeMounts,
-		// TODO: run unprivileged https://github.com/openshift/origin/issues/662
 		corev1.VolumeMount{
-			Name:      "node-storage-root",
-			MountPath: "/var/lib/containers/storage",
+			Name: "node-storage-root",
+			// mounted as additional storage location
+			MountPath: containersAdditionalStoragePath,
+			ReadOnly:  true,
 		},
 	)
 	container.Env = append(container.Env, corev1.EnvVar{Name: "BUILD_STORAGE_DRIVER", Value: "overlay"})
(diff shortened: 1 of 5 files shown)


otaviof:
@adambkaplan, @gabemontero, @nalind, please consider.

gabemontero:
> This pull-request implements the changes described by [Node Cache for Builds enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/builds/node-cache-for-builds.md), bringing read-only access to node images for builds. Please consider the document for full details.
> 
> The modifications were mostly in `setupContainersNodeStorage` and `setupContainersStorage` in order to mount volumes as described bellow.
> ## Volumes and Mounts
> 
> In the following table you can see how the volumes are mounted on build pods.
> Name 	Node (`hostPath`) 	Pod 	Mode
> node-storage-root 	`/var/lib/containers/storage` 	`/var/lib/shared` 	`ro`
> container-storage-root 	None (`EmptyDir{}`) 	`/var/lib/containers/storage` 	`rw`
> 
> The mount location `/var/lib/shared` is chosen due the existing [configuration in openshift/builder](https://github.com/openshift/builder/blob/00f9e90d06c3ac8aeac9642e2af3221eb7f0b6e1/imagecontent/etc/containers/storage.conf#L22). This location could be changed later on, however, that's also the default location for `buildah` upstream.
> ## OpenShift-Builder
> 
> With [current `containers/storage`](https://github.com/openshift/builder/blob/00f9e90d06c3ac8aeac9642e2af3221eb7f0b6e1/go.mod#L10), it will end-up with permission issues on trying to access node layers. Like the following example:
> 
> ```
> error: build error: Failed to push image: error copying layers and metadata from "containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.imagestore=/var/lib/shared]image-registry.openshift-image-registry.svc:5000/otaviof/rails-ex:latest" to "docker://image-registry.openshift-image-registry.svc:5000/otaviof/rails-ex:latest": Error reading blob sha256:2bf094d88b12300b2226c4c24f80533dec869141b52b88f706843329c4f54b1d: error reading blob from source image "containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.imagestore=/var/lib/shared]image-registry.openshift-image-registry.svc:5000/otaviof/rails-ex:latest": not allowed to update mount locations for layers at "/var/run/containers/storage/overlay-layers/mountpoints.json": called a write method on a read-only store
> ```
> 
> Therefore, this pull-request is linked with [PR #164 in openshift-builder](https://github.com/openshift/builder/pull/164).
> ## Testing
> 
> With both images for openshift-controller-manager and openshift-builder (based on [PR #164](https://github.com/openshift/builder/pull/164)), create a new `BuildConfig` object with:
> 
> ```shell
> oc new-app --name=nodejs-ex https://github.com/sclorg/nodejs-ex.git
> ```
> 
> Then, add a distinct label in one of the cluster nodes. For instance:
> 
> ```shell
> oc label nodes ip-10-0-139-172.ec2.internal test-build=true
> ```
> 
> And edit `BuilldConfig` to use the same label:
> 
> ```shell
> oc patch buildconfig/nodejs-ex --patch='{ "spec": { "nodeSelector": { "test-build": "true" } } }'
> ```
> 
> Finally, the `nodejs-ex` build can be started:
> 
> ```shell
> oc start-build nodejs-ex
> ```

So labelling the node/BC .... it might be good to explicitly state here that we do that so the **second** invocation of `oc start-build nodejs-ex` benefits from the cache populated by the **first** invocation.

i.e. we pin **every** invocation of builds for the BC on a particular node.

Later enhancements then could leverage buidlah configuration that allows for shared storage across nodes/hosts, and then we no longer need the node/BC label based pinning.

Probably obvious to our team and those thinking about this for a while, but may not be to outsiders looking in.

adambkaplan:
/hold

As we discussed with @nalind, today there is a non-zero risk that an image layer is pruned by cri-o during the build. If this happens and said layer is not present in the destination registry, image push can fail. This requires a change to buildah's underlying libraries to "pull" the image layers from the additional store so they are reliably stored and can be pushed to the destination registry.

pkg/build/controller/strategy/util.go
```diff
@@ -450,15 +455,16 @@ func setupContainersStorage(pod *corev1.Pod, container *corev1.Container) {
...
+			MountPath: containersStoragePath,
 		},
 	)
 	container.Env = append(container.Env, corev1.EnvVar{Name: "BUILD_STORAGE_DRIVER", Value: "overlay"})
 	container.Env = append(container.Env, corev1.EnvVar{Name: "BUILD_ISOLATION", Value: "chroot"})
 }
 
 // setupContainersNodeStorage borrows the appropriate storage directories from the node so
-// that we can share layers that we're using with the node
+// that we can share layers that we're using with the node, as a additional read-only
```
adambkaplan, line 466: nit
```suggestion
// that we can share layers that we're using with the node, as an additional read-only
```

PR #128, opened 2020-08-11
Bug 1865784: Ingress to route controller: Make setting tls cert based on termination policy

alvaroaleman:
First commit holds a script I used for testing, might be useful for others and is unlikely to cause harm.

I manually verified:
* Passthrough: Route gets created both when the Ingress has and has not a tls cert config 
* Reencrypt: Route gets created both when the Ingress has and has not a tls cert config and has the cert in the former case

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"sync"
 	"time"
 
@@ -468,34 +469,14 @@ func newRouteForIngress(
 	secretLister corelisters.SecretLister,
 	serviceLister corelisters.ServiceLister,
 ) *routev1.Route {
-	var tlsConfig *routev1.TLSConfig
-	if name, ok := referencesSecret(ingress, rule.Host); ok {
-		secret, err := secretLister.Secrets(ingress.Namespace).Get(name)
-		if err != nil {
-			// secret doesn't exist yet, wait
-			return nil
-		}
-		if secret.Type != corev1.SecretTypeTLS {
-			// secret is the wrong type
-			return nil
-		}
-		if _, ok := secret.Data[corev1.TLSCertKey]; !ok {
-			return nil
-		}
-		if _, ok := secret.Data[corev1.TLSPrivateKeyKey]; !ok {
-			return nil
-		}
-		tlsConfig = &routev1.TLSConfig{
-			Termination:                   terminationPolicyForIngress(ingress),
-			Certificate:                   string(secret.Data[corev1.TLSCertKey]),
-			Key:                           string(secret.Data[corev1.TLSPrivateKeyKey]),
-			InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
-		}
-	}
-
 	targetPort, err := targetPortForService(ingress.Namespace, path, serviceLister)
 	if err != nil {
 		// no valid target port
+		return nil
+	}
+
+	tlsSecret, hasInvalidTLSSecret := tlsSecretIfValid(ingress, rule, secretLister)
+	if hasInvalidTLSSecret {
 		return nil
 	}
 
@@ -522,7 +503,7 @@ func newRouteForIngress(
 				Name: path.Backend.ServiceName,
 			},
 			Port: port,
-			TLS:  tlsConfig,
+			TLS:  tlsConfigForIngress(ingress, rule, tlsSecret),
 		},
 	}
 }
@@ -567,17 +548,15 @@ func routeMatchesIngress(
 		return false
 	}
 
-	var secret *corev1.Secret
-	if name, ok := referencesSecret(ingress, rule.Host); ok {
-		secret, _ = secretLister.Secrets(ingress.Namespace).Get(name)
-		if secret == nil {
-			return false
-		}
-	}
-	if !secretMatchesRoute(secret, route.Spec.TLS, terminationPolicyForIngress(ingress)) {
+	tlsSecret, hasInvalidTLSSecret := tlsSecretIfValid(ingress, rule, secretLister)
+	if hasInvalidTLSSecret {
 		return false
 	}
-	return true
+	tlsConfig := tlsConfigForIngress(ingress, rule, tlsSecret)
+	if route.Spec.TLS != nil && tlsConfig != nil {
+		tlsConfig.InsecureEdgeTerminationPolicy = route.Spec.TLS.InsecureEdgeTerminationPolicy
+	}
+	return reflect.DeepEqual(tlsConfig, route.Spec.TLS)
 }
 
 // targetPortForService returns a target port for a Route based on the given
@@ -620,27 +599,6 @@ func targetPortForService(namespace string, path *networkingv1beta1.HTTPIngressP
 		}
 	}
 	return nil, errors.New("no port found")
-}
-
-func secretMatchesRoute(secret *corev1.Secret, tlsConfig *routev1.TLSConfig, terminationPolicy routev1.TLSTerminationType) bool {
-	if secret == nil {
-		return tlsConfig == nil
-	}
-	if secret.Type != corev1.SecretTypeTLS {
-		return tlsConfig == nil
-	}
-	if _, ok := secret.Data[corev1.TLSCertKey]; !ok {
-		return false
-	}
-	if _, ok := secret.Data[corev1.TLSPrivateKeyKey]; !ok {
-		return false
-	}
-	if tlsConfig == nil {
-		return false
-	}
-	return tlsConfig.Termination == terminationPolicy &&
-		tlsConfig.Certificate == string(secret.Data[corev1.TLSCertKey]) &&
-		tlsConfig.Key == string(secret.Data[corev1.TLSPrivateKeyKey])
 }
 
 func splitForPathAndHost(routes []*routev1.Route, host, path string) ([]*routev1.Route, *routev1.Route) {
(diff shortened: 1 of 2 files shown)


alvaroaleman:
For posterity, this is what I used to test:
```
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-reenc-secret
  annotations:
    route.openshift.io/termination: "reencrypt"
spec:
  tls:
  - secretName: tls
    hosts:
    - ingress-reenc-secret
  rules:
  - host: ingress-reenc-secret
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 27443
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-reenc-no-secret
  annotations:
    route.openshift.io/termination: "reencrypt"
spec:
  rules:
  - host: ingress-reenc-no-secret
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 27443
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-passthrough-secret
  annotations:
    route.openshift.io/termination: "passthrough"
spec:
  tls:
  - secretName: tls
    hosts:
    - ingress-passthrough-secret
  rules:
  - host: ingress-passthrough-secret
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 27443
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-passthrough-no-secret
  annotations:
    route.openshift.io/termination: "passthrough"
spec:
  rules:
  - host: ingress-passthrough-no-secret
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 27443
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: ingress-edge-secret
spec:
  tls:
  - secretName: tls
    hosts:
    - ingress-edge-secret
  rules:
  - host: ingress-edge-secret
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 27443
---
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: null
  labels:
    app: test
  name: service-secure
spec:
  ports:
  - name: "27443"
    port: 27443
    protocol: TCP
    targetPort: 27443
  selector:
    app: test
  type: ClusterIP
---
apiVersion: v1
data:
  tls.crt: dGVzdA==
  tls.key: dGVzdA==
kind: Secret
metadata:
  name: tls
  ownerReferences:
type: kubernetes.io/tls
```

```
$ k apply -f ingress.yaml 
ingress.extensions/ingress-reenc-secret created
ingress.extensions/ingress-reenc-no-secret created
ingress.extensions/ingress-passthrough-secret created
ingress.extensions/ingress-passthrough-no-secret created
ingress.extensions/ingress-edge-secret created
service/service-secure configured
secret/tls created


alvaro@t470s-[2020-08-11-16:24]:[admin@openshift-controller-manager][1]/tmp/tmp.h2Dl2cSiij
$ k get route
NAME                                  HOST/PORT                       PATH   SERVICES         PORT    TERMINATION            WILDCARD
ingress-edge-secret-gqx7m             ExtendedValidationFailed               service-secure   27443   edge/Redirect          None
ingress-passthrough-no-secret-dgmv5   ingress-passthrough-no-secret          service-secure   27443   passthrough/Redirect   None
ingress-passthrough-secret-d5q7j      ingress-passthrough-secret             service-secure   27443   passthrough/Redirect   None
ingress-reenc-no-secret-gpgxl         ingress-reenc-no-secret                service-secure   27443   reencrypt/Redirect     None
ingress-reenc-secret-48szc            ExtendedValidationFailed               service-secure   27443   reencrypt/Redirect     None
```

pkg/route/ingress/ingress.go
```diff
@@ -567,14 +548,11 @@ func routeMatchesIngress(
...
 
-	var secret *corev1.Secret
-	if name, ok := referencesSecret(ingress, rule.Host); ok {
-		secret, _ = secretLister.Secrets(ingress.Namespace).Get(name)
-		if secret == nil {
-			return false
-		}
+	tlsConfig, incomplete := tlsConfigForIngress(ingress, rule, tlsSecretIfValid(ingress, rule, secretLister))
+	if route.Spec.TLS != nil && tlsConfig != nil {
+		tlsConfig.InsecureEdgeTerminationPolicy = route.Spec.TLS.InsecureEdgeTerminationPolicy
```
Miciah, line 553: What is the purpose of this assignment?  This prevents the controller resetting the route's `spec.tls.insecureEdgeTerminationPolicy` if the user modifies the generated route, right?  Is that desirable?  None of the other field receive this special treatment.

pkg/route/ingress/ingress.go
```diff
@@ -717,6 +674,64 @@ func generateRouteName(base string) string {
...
 
+func tlsConfigForIngress(
+	ingress *networkingv1beta1.Ingress,
+	rule *networkingv1beta1.IngressRule,
+	potentiallyNilTLSSecret *corev1.Secret,
+) (_ *routev1.TLSConfig, incomplete bool) {
+	if !needsTLSConfig(ingress, rule) {
+		return nil, false
+	}
+	// Edge: must have cert
```
Miciah, line 685: Must this be?  Could we allow edge sans certificate to mean that the ingress uses the ingress controller's default certificate?  (Related: [RFE-339](https://issues.redhat.com/browse/RFE-339) and [BZ1727791](https://bugzilla.redhat.com/show_bug.cgi?id=1727791).)

pkg/route/ingress/ingress.go
```diff
@@ -717,6 +674,64 @@ func generateRouteName(base string) string {
...
+	// Re-Encrypt: May have cert
+	// Passthrough: Must not have cert
+	terminationPolicy := terminationPolicyForIngress(ingress)
+	if potentiallyNilTLSSecret == nil && terminationPolicy == routev1.TLSTerminationEdge {
```
Miciah, line 689: Would it be reasonable to require `spec.tls` to be specified, even if the value is `[{}]`, in order to enable TLS?

pkg/route/ingress/ingress.go
alvaroaleman: There is a pre-existing test that validates this, `TestController_sync/no-op_-_termination_policy_has_been_changed_by_the_user`

pkg/route/ingress/ingress.go
alvaroaleman: I guess that is debatable but I would have argued that if the annotation is set, its implicit that TLS is wanted

Miciah: If it is set to a valid value ("edge", "passthrough", or "reencrypt"), it makes sense that we interpret that as enabling TLS irrespective of `spec.tls`.  If `spec.tls` is omitted and the annotation is set to an unrecognized value (I'm thinking in particular about the empty string or the string "none"), maybe it would make sense to ignore the annotation and not enable TLS.

alvaroaleman: Okay so a valid value on the annotation or an empty element in `spec.tls` (I guess thats supposed to cover the `Allow using TLS without specifying a cert` use-case linked above)?

Miciah: Right, that makes sense to me.

pkg/route/ingress/ingress.go
alvaroaleman: Updated

pkg/route/ingress/ingress.go
alvaroaleman: Updated

pkg/route/ingress/ingress.go
```diff
@@ -717,6 +675,63 @@ func generateRouteName(base string) string {
...
+		tlsConfig.Certificate = string(potentiallyNilTLSSecret.Data[corev1.TLSCertKey])
+		tlsConfig.Key = string(potentiallyNilTLSSecret.Data[corev1.TLSPrivateKeyKey])
+	}
+	return tlsConfig
+}
+
+func tlsEnabled(ingress *networkingv1beta1.Ingress, rule *networkingv1beta1.IngressRule) bool {
+	if val := ingress.Annotations[terminationPolicyAnnotationKey]; val == string(routev1.TLSTerminationPassthrough) || val == string(routev1.TLSTerminationReencrypt) || val == string(routev1.TLSTerminationEdge) {
+		return true
+	}
```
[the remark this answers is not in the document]
alvaroaleman: I don't understand why a switch with a single case is better than an `if`?

pkg/route/ingress/ingress.go
```diff
@@ -468,37 +469,17 @@ func newRouteForIngress(
...
-	}
-
 	targetPort, err := targetPortForService(ingress.Namespace, path, serviceLister)
 	if err != nil {
 		// no valid target port
 		return nil
 	}
 
+	tlsSecret, hasInvalidTLSSecet := tlsSecretIfValid(ingress, rule, secretLister)
+	if hasInvalidTLSSecet {
```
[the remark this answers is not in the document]
alvaroaleman: Done

pkg/route/ingress/ingress.go
```diff
@@ -717,6 +675,63 @@ func generateRouteName(base string) string {
...
+
+func tlsEnabled(ingress *networkingv1beta1.Ingress, rule *networkingv1beta1.IngressRule) bool {
+	if val := ingress.Annotations[terminationPolicyAnnotationKey]; val == string(routev1.TLSTerminationPassthrough) || val == string(routev1.TLSTerminationReencrypt) || val == string(routev1.TLSTerminationEdge) {
+		return true
+	}
+	if len(ingress.Spec.TLS) == 1 && reflect.DeepEqual(ingress.Spec.TLS[0], networkingv1beta1.IngressTLS{}) {
+		return true
+	}
+	_, referencesSecret := referencesSecret(ingress, rule.Host)
+	return referencesSecret
```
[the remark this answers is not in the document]
alvaroaleman: No, because an Ingress can have multiple `IngressRule`s which each lead to a route. In all other cases, we match the `IngressRule` to a `IngressTLS` via its `Host` attribute, the empty one is a special magic way of enabling it without cert for all Ingresses.

pkg/route/ingress/ingress.go
alvaroaleman: Yes that is correct. If someone modified the cert on the route, it would break now. It is however something I intuitively would have expected and required e.G. when using cert manager to update the ingress certs. 

I would hope that no one manually changed a routes cert to something else after it was generated, but I guess Hyums Law applies. Is there a place to call out potentially breaking changes?

pkg/route/ingress/ingress.go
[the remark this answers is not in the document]
Miciah: It's less repetitive, and it's the preferred style, but `if` is acceptable if you want to keep it as-is.

pkg/route/ingress/ingress.go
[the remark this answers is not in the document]
Miciah: Hm, all right.  Is it sufficient to check only the first item in `spec.tls`, or could it make sense for an ingress to specify one rule that has a nonempty TLS config and another rule that uses a second, empty TLS config?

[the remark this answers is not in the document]
alvaroaleman: The empty tls doesn't allow any kind of matching to rules, this is why using it is only allowed if its the only element and it then applies to all of them.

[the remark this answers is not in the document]
Miciah: What about the following ingress?
```yaml
spec:
  tls:
  - secretName: tls
    hosts:
    - foo.com
  - {}
  rules:
  - host: foo.com
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 8443
  - host: foo.apps.mycluster.com
    http:
      paths:
      - backend:
          serviceName: service-secure
          servicePort: 8443
```
I am able to create the ingress, so it seems at least to be valid.  I would expect the above ingress to use the "tls" secret for foo.com and the default certificate for foo.apps.mycluster.com.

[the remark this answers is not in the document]
alvaroaleman: Updated it to accept any empty tlsconfig regardless of what other tlsconfigs are present

alvaroaleman:
Updated the PR to accept an empty tlsconfig anywhere and removed the first commit with the script

alvaroaleman:
@Miciah any chance you can have another look?

pkg/route/ingress/ingress.go
```diff
@@ -717,6 +675,70 @@ func generateRouteName(base string) string {
...
+var emptyTLS = networkingv1beta1.IngressTLS{}
+
+func tlsEnabled(ingress *networkingv1beta1.Ingress, rule *networkingv1beta1.IngressRule) bool {
+	switch ingress.Annotations[terminationPolicyAnnotationKey] {
+	case string(routev1.TLSTerminationPassthrough), string(routev1.TLSTerminationReencrypt), string(routev1.TLSTerminationEdge):
+		return true
+	}
+	if _, referencesSecret := referencesSecret(ingress, rule.Host); referencesSecret {
+		return true
+	}
```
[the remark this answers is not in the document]
alvaroaleman: Done, PTAL

Miciah:
Many thanks!
/lgtm

PR #130, opened 2020-08-18
Bug 1820075: Reflect route status to ingress status

Miciah:
#### ingress: Rename client to routeClient

* `pkg/route/ingress/ingress.go` (`Controller`): Rename "client" to "routeClient".
(`NewController`, `sync`):
* `pkg/route/ingress/ingress_test.go` (`TestController_stabilizeAfterCreate`, `TestController_sync`): Use the new field name.

#### ingress: Rename fake route clientset

* `pkg/route/ingress/ingress_test.go`: Rename import of the fake clientset for the route API to "routev1fake".
(`TestController_stabilizeAfterCreate`, `TestController_sync`): Rename the fake clientset for the route API from "kc" to "routeClientset".  Rename "actions" to "routeActions".

#### ingress: wantCreates, wantPatches, wantDeletes

* `pkg/route/ingress/ingress_test.go` (`TestController_sync`): Rename `wantCreates` to `wantRouteCreates`, `wantPatches` to `wantRoutePatches`, and `wantDeletes` to `wantRouteDeletes`.

#### Reflect route status to ingress status

* `pkg/cmd/controller/ingress.go` (`RunIngressToRouteController`): Pass ingress client to `ingress.NewController`.
* `pkg/route/ingress/ingress.go` (`Controller`): Add `ingressClient` field.
(`NewController`): Add `ingressClient` parameter.  Use the new parameter to set the `Controller` field.
(`sync`): Reflect the canonical hostname from routes that match the ingress to the ingress's status.
* `pkg/route/ingress/ingress_test.go` (`TestController_stabilizeAfterCreate`): Add fake clientset for the networking API.  Use this fake clientset to initialize the controller.
(`TestController_sync`): Add `wantIngressUpdates` to the test inputs.  Add fake clientset for the networking API.  Use `wantIngressUpdates` in new test cases for ingress status updates.  Use the fake networking clientset to initialize the controller and to verify the new tests cases.

----

Depends on https://github.com/openshift/openshift-apiserver/pull/129 and https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/169.

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -26,6 +26,7 @@ import (
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	networkingv1beta1informers "k8s.io/client-go/informers/networking/v1beta1"
 	kv1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	kv1beta1networking "k8s.io/client-go/kubernetes/typed/networking/v1beta1"
 	corelisters "k8s.io/client-go/listers/core/v1"
 	networkingv1beta1listers "k8s.io/client-go/listers/networking/v1beta1"
 	"k8s.io/client-go/tools/cache"
@@ -74,7 +75,8 @@ import (
 type Controller struct {
 	eventRecorder record.EventRecorder
 
-	client routeclient.RoutesGetter
+	routeClient   routeclient.RoutesGetter
+	ingressClient kv1beta1networking.IngressesGetter
 
 	ingressLister networkingv1beta1listers.IngressLister
 	secretLister  corelisters.SecretLister
@@ -161,7 +163,7 @@ type queueKey struct {
 }
 
 // NewController instantiates a Controller
-func NewController(eventsClient kv1core.EventsGetter, client routeclient.RoutesGetter, ingresses networkingv1beta1informers.IngressInformer, secrets coreinformers.SecretInformer, services coreinformers.ServiceInformer, routes routeinformers.RouteInformer) *Controller {
+func NewController(eventsClient kv1core.EventsGetter, routeClient routeclient.RoutesGetter, ingressClient kv1beta1networking.IngressesGetter, ingresses networkingv1beta1informers.IngressInformer, secrets coreinformers.SecretInformer, services coreinformers.ServiceInformer, routes routeinformers.RouteInformer) *Controller {
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartLogging(klog.Infof)
 	// TODO: remove the wrapper when every clients have moved to use the clientset.
@@ -175,7 +177,8 @@ func NewController(eventsClient kv1core.EventsGetter, client routeclient.RoutesG
 		expectations:     newExpectations(),
 		expectationDelay: 2 * time.Second,
 
-		client: client,
+		routeClient:   routeClient,
+		ingressClient: ingressClient,
 
 		ingressLister: ingresses.Lister(),
 		secretLister:  secrets.Lister(),
@@ -383,7 +386,7 @@ func (c *Controller) sync(key queueKey) error {
 
 	// walk the ingress and identify whether any of the child routes need to be updated, deleted,
 	// or created, as efficiently as possible.
-	var creates, updates []*routev1.Route
+	var creates, updates, matches []*routev1.Route
 	for _, rule := range ingress.Spec.Rules {
 		if rule.HTTP == nil {
 			continue
@@ -406,6 +409,7 @@ func (c *Controller) sync(key queueKey) error {
 			}
 
 			if routeMatchesIngress(existing, ingress, &rule, &path, c.secretLister, c.serviceLister) {
+				matches = append(matches, existing)
 				continue
 			}
 
@@ -424,7 +428,7 @@ func (c *Controller) sync(key queueKey) error {
 
 	// add the new routes
 	for _, route := range creates {
-		if err := createRouteWithName(c.client, ingress, route, c.expectations); err != nil {
+		if err := createRouteWithName(c.routeClient, ingress, route, c.expectations); err != nil {
 			errs = append(errs, err)
 		}
 	}
@@ -436,7 +440,7 @@ func (c *Controller) sync(key queueKey) error {
 			return err
 		}
 		data = []byte(fmt.Sprintf(`[{"op":"replace","path":"/spec","value":%s}]`, data))
-		_, err = c.client.Routes(route.Namespace).Patch(context.TODO(), route.Name, types.JSONPatchType, data, metav1.PatchOptions{})
+		_, err = c.routeClient.Routes(route.Namespace).Patch(context.TODO(), route.Name, types.JSONPatchType, data, metav1.PatchOptions{})
 		if err != nil {
 			errs = append(errs, err)
 		}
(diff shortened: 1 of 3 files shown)


pkg/route/ingress/ingress.go
```diff
@@ -435,15 +439,45 @@ func (c *Controller) sync(key queueKey) error {
...
 
 	// purge any previously managed routes
 	for _, route := range old {
-		if err := c.client.Routes(route.Namespace).Delete(context.TODO(), route.Name, metav1.DeleteOptions{}); err != nil && !kerrors.IsNotFound(err) {
+		if err := c.routeClient.Routes(route.Namespace).Delete(context.TODO(), route.Name, metav1.DeleteOptions{}); err != nil && !kerrors.IsNotFound(err) {
+			errs = append(errs, err)
+		}
+	}
+
+	// reflect route status to ingress status
```
sgreene570, line 455: If I'm following these changes correctly, route status should only be reflected back to the parent ingress object if `len(matches) > 0`, right? If that's true, would it be worth encasing lines 456->483 in an if statement that checks for the length of `matches`?

Miciah: I don't think that's right.  What if the ingress is created in such a way that corresponding routes are created and their statuses reflected back to the ingress, and then the ingress is modified in such a way that the routes are removed?  The status should be cleared in that case.  Hm, maybe this needs a test case...

sgreene570: Ah, I see,I missed that edge case.

In terms of adding a test case, perhaps one of the `delete route when` test cases could be enhanced to verify that Ingress status conditions are updated properly when corresponding routes are deleted? The current code coverage for the changes in this PR is fantastic, but I think this potential flow should be visible from the test cases (or at least some in-line comments).

adambkaplan:
/hold

Shouldn't this require https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/169 to merge first?
Do we have an e2e test which ensures the ingress-to-route controller successfully updates status?

Miciah:
> Shouldn't this require [openshift/cluster-openshift-controller-manager-operator#169](https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/169) to merge first?

Yeah, that has been the piece blocking this PR.

> Do we have an e2e test which ensures the ingress-to-route controller successfully updates status?

No, just unit tests.

pkg/route/ingress/ingress.go
Miciah: Done.  Thanks!

pkg/route/ingress/ingress_test.go
```diff
@@ -463,6 +469,7 @@ func TestController_sync(t *testing.T) {
 		wantCreates     []*routev1.Route
 		wantPatches     []clientgotesting.PatchActionImpl
 		wantDeletes     []clientgotesting.DeleteActionImpl
+		wantUpdates     []clientgotesting.UpdateActionImpl
```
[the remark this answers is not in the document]
Miciah: Done.  I also renamed `wantCreates` to `wantRouteCreates`, `wantPatches` to `wantRoutePatches`, and `wantDeletes` to `wantRouteDeletes`.

Miciah:
Rebased.

Miciah:
The failure apparently wasn't related to ingress-to-route this time:

    FAIL: github.com/openshift/openshift-controller-manager/pkg/serviceaccounts/controllers TestUpdateNewStyleSecretIPv6 45s

/test unit

Miciah:
`TestUpdateNewStyleSecretIPv6` failed again.  Unit tests all pass locally.  One more try, and if it fails again, I'll open a test PR.  If the test PR fails, I'll open a Bugzilla report.
/test unit

Miciah:
H, this time `TestUpdateOldStyleSecretWithoutKey` failed.

Miciah:
Unit tests passed on #135.
/test unit

Miciah:
And now unit tests are passing here.

Miciah:
openshift/cluster-openshift-controller-manager-operator#169 merged, so
/hold cancel

adambkaplan (commented):
@Miciah squash commits, and then I think this looks good. I'd like someone from the network-edge team to LGTM before merge.

/approve

Miciah:
> @Miciah squash commits, and then I think this looks good. I'd like someone from the network-edge team to LGTM before merge.

Done (but I think it hurts readability).  Thanks!  @sgreene570, do you mind giving this another look?

sgreene570:
@Miciah @adambkaplan I reviewed these changes again and have no further comments :grin: 
/lgtm

sgreene570:
@Miciah this PR seems to be consistently failing the `: [sig-network] Ingress API should support creating Ingress API operations [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s] ` test.

Miciah:
> @Miciah this PR seems to be consistently failing the `[sig-network] Ingress API should support creating Ingress API operations [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]` test in the e2e-aws job.

The test creates an ingress, futzes with it a bit, and then tries to patch its status:  
https://github.com/kubernetes/kubernetes/blob/b4ffb455315c/test/e2e/network/ingress.go#L977-L1143
But now the ingress-to-route controller is setting the ingress's status.  The immediate problem is that the test doesn't get before patching, so it has a stale copy, so the update fails.  The bigger problem is that the test assumes that nothing else updates the ingress's status.  Because the test does not set `spec.ingressClassName`, such an expectation is _almost_ reasonable, but the API godoc says, "Implementations of this API should ignore Ingresses without a class specified."  Not _"must"_, just _"should"_.  So strictly speaking, ingress-to-route isn't totally nonconformant here.  This all raises some questions:  
* Is the test's expectation that nothing else updates the ingress's status reasonable?  I'd argue it isn't (see "not _'must'_, just _'should'_").  
* Does the ingress-to-route controller need to start respecting `ingressClassName`?  We'd need (a) to invent a class name, (b) to start setting the default class, (c) to add the class to existing ingresses, and (d) to start respecting the class in the ingress-to-route controller.  It's a good bit of effort (especially migrating existing ingresses) but probably a good long-term goal.  
* Should we skip the test for now?  Updating the status doesn't make us any less conformant as we're already _not_ ignoring the ingress, and I'd argue that updating the status makes us _more_ conformant as long as we _aren't_ ignoring the ingress because the godoc of `status.loadBalancer` is, "LoadBalancer contains the current status of the load-balancer."  So we are nonconformant if we expose the ingress as a route and fail to reflect that to the ingress's status.

Thoughts?

Miciah:
/hold
while we debate API semantics.

Miciah:
Another question:  Why wasn't this test failing before?

Miciah:
I suppose one could make the case that an ingress controller must not remove entries from `status.loadBalancer` that the controller did not itself add, because an ingress controller cannot know that it alone exposes  the ingress.  By that line of reasoning, the test is still broken, but ingress-to-route should be more careful not to clear out entries that it did not add.

sgreene570:
> * Should we skip the test for now?  Updating the status doesn't make us any less conformant as we're already _not_ ignoring the ingress, and I'd argue that updating the status makes us _more_ conformant as long as we _aren't_ ignoring the ingress because the godoc of `status.loadBalancer` is, "LoadBalancer contains the current status of the load-balancer."  So we are nonconformant if we expose the ingress as a route and fail to reflect that to the ingress's status.

Yea, I think in this case it would make sense to temporarily skip the test and revisit it later. To your first point, I would also argue that it is unreasonable for the test to expect _no_ changes to the Ingress status post Ingress creation.

> Does the ingress-to-route controller need to start respecting ingressClassName? 

Most likely yes. I agree this is a non-trivial change, but is still a good long term goal.

> By that line of reasoning, the test is still broken, but ingress-to-route should be more careful not to clear out entries that it did not add.

Does this warrant changes to this PR, or are you just making a general note about how the ingress-to-route controller should work?

Miciah:
> > By that line of reasoning, the test is still broken, but ingress-to-route should be more careful not to clear out entries that it did not add.
> 
> Does this warrant changes to this PR, or are you just making a general note about how the ingress-to-route controller should work going forward?

I went ahead and pushed a change to that effect.

sgreene570:
New changes look good to me. 
@Miciah ping me once the test in question passes or is skipped for `lgtm`

sgreene570:
hm, seconds after I left the above comment the e2e job passed.
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_openshift-controller-manager/130/pull-ci-openshift-openshift-controller-manager-master-e2e-aws/1308798128467480576 passed all tests :thinking: 

@Miciah you were under the impression that this PR should continue to fail on `[sig-network] Ingress API should support creating Ingress API operations [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]`?

Miciah:
> hm, seconds after I left the above comment the e2e job passed.
> https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_openshift-controller-manager/130/pull-ci-openshift-openshift-controller-manager-master-e2e-aws/1308798128467480576 passed all tests 
> 
> @Miciah you were under the impression that this PR should continue to fail on `[sig-network] Ingress API should support creating Ingress API operations [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]`?

I wasn't sure.  It would depend on whether the problem was that ingress-to-route was adding a status entry or whether it was because ingress-to-route was deleting the test's status entry.  If the test is passing now, it must have been the latter (as that is the behavior that I just changed).  Let's rerun the test to be sure though.

/test e2e-aws

sgreene570:
Test has passed again! :partying_face: 
/lgtm

sgreene570:
@Miciah is the hold on this PR still necessary?

Miciah:
We can cancel the hold.  Thanks!
/hold cancel

PR #136, opened 2020-09-22
WIP: Bug 1876500: buildConfig lastTriggeredImageID needs to be maintained in buildConfig.status

gabemontero:
has place holder API update commit until https://github.com/openshift/api/pull/744 is merged, then this PR will vendor in that change, and the place holder commit will be removed from this PR

/assign @adambkaplan

### DIFF
--- a/pkg/image/controller/trigger/image_trigger_controller_test.go
+++ b/pkg/image/controller/trigger/image_trigger_controller_test.go
@@ -596,6 +596,11 @@ func scenario_1_buildConfig_strategy() *buildv1.BuildConfig {
 				},
 			},
 		},
+		Status: buildv1.BuildConfigStatus{
+			ImageChangeTriggersState: []buildv1.ImageChangeTrigger{
+				{},
+			},
+		},
 	}
 }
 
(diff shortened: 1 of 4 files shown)


gabemontero:
The image change trigger tests in e2e-aws are failing as expected since the api server changes in https://github.com/openshift/openshift-apiserver/pull/138 are needed for them to pass @adambkaplan 

what I'm now wondering is whether we need both OCM and apiserver changes in for that test to pass, and if we have to put in an intermediate form that looks at both spec and status 

will have a better sense once the tests in https://github.com/openshift/openshift-apiserver/pull/138 complete

gabemontero:
ok @adambkaplan the pruning tests are passing again with commit https://github.com/openshift/openshift-controller-manager/pull/136/commits/7562bac0177c250570bb60e815d46d262651b9a2 where we leave the unit test changes as is, where they validate the status LastTriggeredImageID is set, but set both spec and status in the actualy code, to correlate with an openshift-apiserver implementation that only deals with spec.

Similar change cooking on the openshift-apiserver side

So when 4.7 opens up and we can merge https://github.com/openshift/api/pull/744, vendor that here with this PR, we can pass e2e's, and then follow up with a subsequent PR that removes https://github.com/openshift/openshift-controller-manager/pull/136/commits/7562bac0177c250570bb60e815d46d262651b9a2

gabemontero:
/close

between @pweil- @bparees @adambkaplan and myself we've concluded removal of this from spec could break existing users who built on the current behavior

@adambkaplan is driving mitigations with customer ... based on fallout, if some altered form of this is taken on, we'll go from there

PR #137, opened 2020-09-30
Bug 1862532: Fix label value validation for Builds

coreydaley-redhat:
Label check should allow underscores, hyphens, and decimals in the name
instead of restricting to DNS RFC 1123 compatible values


adambkaplan:
/retitle Bug 1862532: Fix label value validation for Builds

adambkaplan (approved):
/lgtm

PR #142, opened 2020-10-28
[WIP] initial code for JIRA ticket

coreydaley:


coreydaley-redhat:
@nalind Do you have any thoughts on resolving this error for this pull request?

`2020-10-28T16:28:33.861510089Z F1028 16:28:33.861486       1 helpers.go:115] error: vfs driver does not support overlay.imagestore options`

nalind:
The `overlay.imagestore` storage options must be coming from the /etc/containers/storage.conf in the builder container.  I haven't personally verified this (lately?), but if we set the `BUILD_STORAGE_OPTIONS` environment variable to a JSON-encoded slice of strings (i.e., `[]`), that should override the value from the configuration file.

Failing that, arranging for an alternative version of storage.conf to show up in the container, and setting `BUILD_STORAGE_CONF_PATH` to point to it should let you set everything that way.

PR #152, opened 2020-12-19
ingress-to-route: add support for dest ca cert

rcgoodfellow:
This PR adds support for `route.openshift.io/v1/Route.spec.tls.destinationCACertificate` [[1](https://docs.openshift.com/container-platform/4.6/rest_api/network_apis/route-route-openshift-io-v1.html)] as the following Ingress Annotation

```
route.openshift.io/destinationCACertificate
```

This is needed to support reencrypt termination where the underlying pod has a self signed cert [[2](https://access.redhat.com/solutions/4601031)].

With this PR the following works

```yaml
kind: Ingress
metadata:
  annotations:
    route.openshift.io/destinationCACertificate: |
      -----BEGIN CERTIFICATE-----
      <snip>
      -----END CERTIFICATE-----
```

I've tested this in the latest Code Ready Containers (v1.20 at time of writing), which has an OpenShift cluster version of 4.6.6. The code modifications were tested using a cherry pick of the changes in this PR onto a [release-4.6 based branch](https://github.com/mergetb/openshift-controller-manager/tree/release-4.6-ingress-destination-ca).

The change to the Dockerfile is to make it possible for folks outside of RH to build the container. If this is not desired, or there is a better publicly accessible alternative I'm happy to remove or update that change.

[1] https://docs.openshift.com/container-platform/4.6/rest_api/network_apis/route-route-openshift-io-v1.htm
[2] https://access.redhat.com/solutions/4601031

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.13 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
 WORKDIR /go/src/github.com/openshift/openshift-controller-manager
 COPY . .
 RUN make build --warn-undefined-variables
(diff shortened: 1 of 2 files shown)


rcgoodfellow:
Sorry this came from the wrong branch, closing and reopening.

PR #153, opened 2020-12-20
ingress-to-route: add support for dest ca cert

rcgoodfellow:
This PR adds support for `route.openshift.io/v1/Route.spec.tls.destinationCACertificate` [[1](https://docs.openshift.com/container-platform/4.6/rest_api/network_apis/route-route-openshift-io-v1.html)] as the following Ingress Annotation

```
route.openshift.io/destinationCACertificate
```

This is needed to support reencrypt termination where the underlying pod has a self signed cert [[2](https://access.redhat.com/solutions/4601031)].

With this PR the following works

```yaml
kind: Ingress
metadata:
  annotations:
    route.openshift.io/destinationCACertificate: |
      -----BEGIN CERTIFICATE-----
      <snip>
      -----END CERTIFICATE-----
```

I've tested this in the latest Code Ready Containers (v1.20 at time of writing), which has an OpenShift cluster version of 4.6.6. The code modifications were tested using a cherry pick of the changes in this PR onto a [release-4.6 based branch](https://github.com/mergetb/openshift-controller-manager/tree/release-4.6-ingress-destination-ca).

The change to the Dockerfile is to make it possible for folks outside of RH to build the container. If this is not desired, or there is a better publicly accessible alternative I'm happy to remove or update that change.

[1] https://docs.openshift.com/container-platform/4.6/rest_api/network_apis/route-route-openshift-io-v1.htm
[2] https://access.redhat.com/solutions/4601031

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.14 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.14 AS builder
 WORKDIR /go/src/github.com/openshift/openshift-controller-manager
 COPY . .
 RUN make build --warn-undefined-variables
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.14 AS builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.14 AS builder
 WORKDIR /go/src/github.com/openshift/openshift-controller-manager
 COPY . .
 RUN make build --warn-undefined-variables
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/openshift-controller-manager/openshift-controller-manager /usr/bin/
 LABEL io.k8s.display-name="OpenShift Controller Manager Command" \
       io.k8s.description="OpenShift is a platform for developing, building, and deploying containerized applications." \
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -747,9 +747,11 @@ func tlsConfigForIngress(
 	// Re-Encrypt: May have cert
 	// Passthrough: Must not have cert
 	terminationPolicy := terminationPolicyForIngress(ingress)
+	destinationCACertificate := destinationCACertificateForIngress(ingress)
 	tlsConfig := &routev1.TLSConfig{
 		Termination:                   terminationPolicy,
 		InsecureEdgeTerminationPolicy: routev1.InsecureEdgeTerminationPolicyRedirect,
+		DestinationCACertificate:      destinationCACertificate,
 	}
 	if terminationPolicy != routev1.TLSTerminationPassthrough && potentiallyNilTLSSecret != nil {
 		tlsConfig.Certificate = string(potentiallyNilTLSSecret.Data[corev1.TLSCertKey])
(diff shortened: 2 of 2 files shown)


rcgoodfellow:
@soltysh @knobunc thoughts on this?

glawler:
@soltysh @knobunc Any updates on this PR? Thanks.

frobware:
I spent a little time trying to understand the issue/failure/ask here. It's not clear to me how one destination CA can be _always_ applicable for the whole of one ingress controller.

The KB article you linked to mentions you will get a 503 for reencrypt routes.
I played around here[1] and could not reproduce the 503 at least. Any chance you could look at my experiment to see if I'm doing something there which would mitigate your PR?

[1] - https://github.com/frobware/haproxy-hacks/tree/master/destca

davidkarlsen:
> I spent a little time trying to understand the issue/failure/ask here. It's not clear to me how one destination CA can _always_ be applicable for the whole of one ingress controller.
It's not for the whole ingress controller, but a specific ingress resource, which then openshift picks up and creates a route for. In most cases this will be a single service, or services with the same TLS ca.

This PR makes a lot of sense to me.

dabelenda:
Am I the only one thinking that setting the annotation `route.openshift.io/termination` to `reencrypt` without a way to set `destinationCACertificate` in the resulting Route is nearly useless? There is no way to make the HAProxy ignore an invalid backend certificate using annotations either so the Route will always return an error except when it is signed with the kubernetes Service CA.

In the current code the `destinationCACertificate` is preserved if the Route already has one, so you need to manually patch the randomly-named route after it is created to add the CA. I guess the only choice is to use cert-utils-operator to have an automated way to add the CA in the Route after it was created. I wonder why this feature was not added immediately when the annotation "route.openshift.io/termination" was created.

To reproduce the issue we are talking about:

Create a CA certificate, for example:
```
$ export MY_FQDN=test.example.com # change this to the FQDN you want in your certificate
$ mkdir test-ca-temp
$ cd test-ca-temp
$ openssl req -x509 -new -nodes -newkey rsa:2048 -keyout ca.key -sha256 -days 1825 -out ca.crt -subj '/CN=My Test CA/' 
$ openssl req -new -newkey rsa:2048 -nodes -keyout server.key -out server.csr -subj "/CN=$MY_FQDN/"
$ cat << EOF > ca.conf           
[ certificate ]                                                                                                                             

subjectAltName = @alt_names
authorityKeyIdentifier=keyid,issuer:always
basicConstraints=CA:FALSE
keyUsage=keyEncipherment,dataEncipherment
extendedKeyUsage=serverAuth,clientAuth

[ alt_names ]
DNS.1 = $MY_FQDN
EOF
$ openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 10000 -extfile ca.conf -extensions certificate
```

With the above you get a CA certificate in `ca.crt` a TLS certificate in `server.crt` valid for `$MY_FQDN` (you most likely want to change that) and the key in `server.key`. Then you can deploy a tiny application serve HTTPS using this certificate and try to create an ingress for that application. And see that the openshift router will always send a "application not available" until you set the destinationCACertificate in the Route spec.

RyuunoAelia:
@frobware https://github.com/frobware/haproxy-hacks/blob/master/destca/destca-deployment.yaml#L6 This is why you cannot reproduce, the CA that openshift will use to populate your secret is implicitely trusted by HAProxy in the OpenShift ingress controller setup. If you want to get the same issue as us, you need another CA for your application's TLS.

YentlFrickx:
I agree with @dabelenda , we currently can't use Ingress objects because we can't use reencrypt (which our use case specifically needs)

glawler:
So what's going on with this?

ctml91:
> Am I the only one thinking that setting the annotation `route.openshift.io/termination` to `reencrypt` without a way to set `destinationCACertificate` in the resulting Route is nearly useless?

Nope, I am 100% with you and it kind of boggles my mind that it isn't supported yet. So basically, we are SOL... lol

Miciah:
Thanks for the PR!  #218 has merged and implements the feature with the changes I had suggested in my earlier comment, so we can close this PR.  
/close

PR #156, opened 2021-01-12
r&d on fall out from turning off privileged

gabemontero:
@bparees @nalind per our discussion today here is the r & d foray on what breaks in buildah when we turn off the privileged bit in the build pod

I'll be diagnosing the test failures and report back any details as I find them.

I'll set up cluster(s) with this change for any exploration as well.

### DIFF
--- a/pkg/build/controller/strategy/docker.go
+++ b/pkg/build/controller/strategy/docker.go
@@ -90,7 +90,7 @@ func (bs *DockerBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 							MountPath: buildutil.BuildBlobsMetaCache,
 						},
 					},
-					ImagePullPolicy: v1.PullIfNotPresent,
+					ImagePullPolicy: v1.PullAlways,
 					Resources:       build.Spec.Resources,
 				},
 			},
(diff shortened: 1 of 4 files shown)


gabemontero:
ton of `time="2021-01-12T17:41:54Z" level=error msg="'overlay' is ...ported over xfs at \"/var/lib/containers/storage/overlay\""
		error: kernel does not support overlay fs: 'overlay' is no...": backing file system is unsupported for this graph driver` as expected with overlay on

seeing if we can quickly switch to vfs

gabemontero:
@bparees @nalind I've pushed the changes to 
- have the OCM daemonset alway pull the container image to facilitate iterative development on the OCM by updating any dev images used (atm this fails a unit test)
-  after moving privileged to false with this PR's initial commit, the other commit I just pushed now sets the storage driver to vfs (like we do in tekton when we run unprivileged)

cluster-bot could be used to launch clusters with these changes if you want to try them out on your own

Builds still fail with these changes, in a way that did not occur when we did the analogous change in tekton, which point to permission profile diffs and seccomp related diffs between the two envs.  If you'd like the gory details @bparees check out the discussion thread you, @nalind and I were on today.

pkg/build/controller/strategy/docker.go
```diff
@@ -90,7 +90,7 @@ func (bs *DockerBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 							MountPath: buildutil.BuildBlobsMetaCache,
 						},
 					},
-					ImagePullPolicy: v1.PullIfNotPresent,
```
gabemontero, line 93: only did this for iterative development ... not needed to go unprivieged

PR #160, opened 2021-01-19
remove use of BUILD_ISOLATION env var (no longer inspected in openshift/builder)

gabemontero:
/assign @bparees 

@nalind @jupierce


gabemontero:
fyi i think I started this before build 206 actually merged so I'm anticipating needing a retest

PR #162, opened 2021-01-22
Bug 1911470: Creating secrets do access registry through routes

ricardomaraschini:
This commits adds registry external routes to created docker
configuration secrets.

It hooks the controller manager into Image config changes.

### DIFF
--- a/pkg/cmd/controller/serviceaccount.go
+++ b/pkg/cmd/controller/serviceaccount.go
@@ -78,6 +78,7 @@ func RunServiceAccountPullSecretsController(ctx *ControllerContext) (bool, error
 	go serviceaccountcontrollers.NewDockerRegistryServiceController(
 		ctx.KubernetesInformers.Core().V1().Secrets(),
 		ctx.KubernetesInformers.Core().V1().Services(),
+		ctx.ConfigInformers.Config().V1().Images(),
 		kc,
 		dockerRegistryControllerOptions,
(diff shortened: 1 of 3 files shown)


dmage:
AFAIK all OCM configuration parameters should be wired through its operator. Otherwise it'll be more difficult to investigate problems as we won't know the controller state.

adambkaplan:
I agree with @dmage here - the better approach is to have openshift-controller-manager-operator wire the external hostnames in via the mounted OCM configuration file.

It seems that the pull secrets controller already has a configuration API [1]. I'm not 100% clear on how `RegistryURLs` is used - we could either take advantage of that field or create a new field for the external URLs.

[1] https://github.com/openshift/api/blob/a9e731090f5ed361e5ab887d0ccd55c1db7fc633/openshiftcontrolplane/v1/types.go#L232

adambkaplan:
We already wire in the internal registry hostname here - https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/master/pkg/operator/configobservation/images/observe_images.go#L15-L54

ricardomaraschini:
@adambkaplan I have checked the code and seems like RegistryURLs is not in use. I am closing this PR and opening a new one with the patch to start using this config directive. PTAL at https://github.com/openshift/cluster-openshift-controller-manager-operator/pull/197

PR #165, opened 2021-01-28
Bug 1900989: unidling: switch away from endpoints to the service

frobware:
Previously idle annotations were added to endpoints. In 4.6 the router
switched to watching endpointslices which don't have the idle
annotation. There was some upstream PRs/discussion to automatically
propagate existing annotations from endpoints to endpointslices:

  https://github.com/kubernetes/kubernetes/pull/98116

As that was a non-starter we changed oc/idle to annotate the service
(and the endpoints for backwards compatibility). In the router we are
already watching service and and when we get endpoint changes we now
check for the idle annotation on the service. If it is present we
write the cluster-ip address into the haproxy.config as we previously
did prior to switching to endpointslices. We watch endpointslices for 
dual-stack and IPv6 support.

This change means that undiling needs to watch for services with idle
annotations as opposed to endpoints. Switch the unidling controller to 
look for idle annotations on the service.

Requires:
- https://github.com/openshift/oc/pull/720
- https://github.com/openshift/router/pull/225
- ~https://github.com/openshift/sdn/pull/252~
- https://github.com/openshift/openshift-apiserver/pull/180
- https://github.com/openshift/origin/pull/25844
- https://github.com/openshift/origin/pull/25847

TODO: amend  origin test (https://github.com/openshift/origin/pull/25847)  to exercise idle/unidle via a route and not just the service.

### DIFF
--- a/pkg/cmd/controller/unidling.go
+++ b/pkg/cmd/controller/unidling.go
@@ -34,6 +34,7 @@ func RunUnidlingController(ctx *ControllerContext) (bool, error) {
 		ctx.RestMapper,
 		coreClient,
 		coreClient,
+		coreClient,
 		appsClient.AppsV1(),
 		coreClient,
 		resyncPeriod,
--- a/pkg/unidling/controller/unidling_controller.go
+++ b/pkg/unidling/controller/unidling_controller.go
@@ -66,19 +66,19 @@ func (c *lastFiredCache) AddIfNewer(info types.NamespacedName, newLastFired time
 }
 
 type UnidlingController struct {
-	controller          cache.Controller
-	scaleNamespacer     scale.ScalesGetter
-	mapper              meta.RESTMapper
-	endpointsNamespacer corev1client.EndpointsGetter
-	queue               workqueue.RateLimitingInterface
-	lastFiredCache      *lastFiredCache
+	controller         cache.Controller
+	scaleNamespacer    scale.ScalesGetter
+	mapper             meta.RESTMapper
+	servicesNamespacer corev1client.ServicesGetter
+	queue              workqueue.RateLimitingInterface
+	lastFiredCache     *lastFiredCache
 
 	// TODO: remove these once we get the scale-source functionality in the scale endpoints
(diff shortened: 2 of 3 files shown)


frobware:
Requires:

- openshift/oc#720
- openshift/router#225
- openshift/sdn#252

frobware:
cc @openshift/openshift-team-network-edge

frobware:
Also needs https://github.com/openshift/openshift-apiserver/pull/180

sttts:
Why didn't we notice that idling got broken? Is there no e2e test? Does one of the PRs add a test?

frobware:
> Why didn't we notice that idling got broken? Is there no e2e test? Does one of the PRs add a test?

We have an idling test that is currently excluded; PR to re-enable https://github.com/openshift/origin/pull/25847.
However, it turns out that that idling test didn't exercise idling through a router.

soltysh (approved):
/lgtm
/approve

danwinship:
> This change means that unidling needs to watch for services with idle
> annotations as opposed to endpoints.

Why does it mean that? `oc idle` was changed to annotate _both_, therefore existing components should be able to continue watching endpoints for now, shouldn't they?

danwinship:
This change would also mean that if you tried to use a 4.6 `oc` to idle services in a 4.7 cluster, it would claim to be doing something, but wouldn't actually have any effect.

soltysh:
> This change would also mean that if you tried to use a 4.6 `oc` to idle services in a 4.7 cluster, it would claim to be doing something, but wouldn't actually have any effect.

Correct, that's why we should ensure the oc bits gets backported asap to 4.6

frobware:
> Why does it mean that? `oc idle` was changed to annotate _both_, therefore existing components should be able to continue watching endpoints for now, shouldn't they?

Yes, this is true. Uniding should remove idle annotations from services.

frobware:
I think we can simplify this. The additions here should remove idle annotations on services.

frobware:
https://github.com/openshift/oc/pull/720 has merged.

/hold cancel

PR #173, opened 2021-03-17
BUILD-433: Run builds in user namespaces without seccomp if BUILD_PRIVILEGED=false

nalind:
In the build controller, we never actually called `setupContainersNodeStorage()`, so remove it, and stop setting the environment variable to force storage to use the overlay driver, which currently requires privileges, leaning on https://github.com/openshift/builder/pull/220 and https://github.com/openshift/builder/pull/291 to have the builder container figure that out for itself.

~Make the mode which we apply to secret and configuration map volumes in build pods something that the build controller can pass down to its helper functions, so that it can be adjusted based on the strategy.  Continue to default them to 0600.~

Set the seccomp profile for build pods to "unconfined", since we depend on `unshare()` and CRI-O is attempting to move away from allowing it to be used in its default seccomp profile.

When the Build object's environment includes "BUILD_PRIVILEGED=false" or "BUILD_PRIVILEGED=0":
* turn off the "privileged" flag for build pods
* set "io.openshift.builder=" and "io.kubernetes.cri-o.Devices=/dev/fuse:rwm" annotations
* set "io.kubernetes.cri-o.userns-mode=auto:size=65536"

### DIFF
--- a/pkg/build/controller/strategy/custom.go
+++ b/pkg/build/controller/strategy/custom.go
@@ -88,7 +88,7 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 		serviceAccount = buildutil.BuilderServiceAccountName
 	}
 
-	privileged := true
+	securityContext := securityContextForBuild(strategy.Env)
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      buildutil.GetBuildPodName(build),
@@ -99,13 +99,10 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 			ServiceAccountName: serviceAccount,
 			Containers: []corev1.Container{
 				{
-					Name:  CustomBuild,
-					Image: strategy.From.Name,
-					Env:   containerEnv,
-					// TODO: run unprivileged https://github.com/openshift/origin/issues/662
-					SecurityContext: &corev1.SecurityContext{
-						Privileged: &privileged,
-					},
+					Name:                     CustomBuild,
+					Image:                    strategy.From.Name,
+					Env:                      containerEnv,
+					SecurityContext:          securityContext,
 					TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
 				},
 			},
@@ -138,6 +135,10 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 	setupAdditionalSecrets(pod, &pod.Spec.Containers[0], build.Spec.Strategy.CustomStrategy.Secrets)
 	setupContainersConfigs(build, pod)
 	setupBuildCAs(build, pod, additionalCAs, internalRegistryHost)
-	setupContainersStorage(pod, &pod.Spec.Containers[0]) // for unprivileged builds
+	setupContainersStorage(pod, &pod.Spec.Containers[0])
+	if securityContext == nil || securityContext.Privileged == nil || !*securityContext.Privileged {
+		setupBuilderAutonsUser(build, strategy.Env, pod)
+		setupBuilderDeviceFUSE(pod)
+	}
 	return pod, nil
 }
--- a/pkg/build/controller/strategy/util.go
+++ b/pkg/build/controller/strategy/util.go
@@ -148,7 +148,10 @@ func mountVolume(pod *corev1.Pod, container *corev1.Container, objName, mountPat
 			break
 		}
 	}
-	mode := int32(0600)
+	mode := int32(0o600)
+	if container.SecurityContext == nil || container.SecurityContext.Privileged == nil || !*container.SecurityContext.Privileged {
+		mode = int32(0o644) // make sure unprivileged builders can read them
+	}
 	if !volumeExists {
 		volume := makeVolume(volumeName, objName, mode, fsType, volumeSource)
 		pod.Spec.Volumes = append(pod.Spec.Volumes, volume)
@@ -465,17 +468,19 @@ func updateConfigsForContainer(c corev1.Container, volumeName string, configDir 
 	return c
 }
 
-// setupContainersStorage creates a volume that we'll use for holding images and working
+// setupContainersStorage creates volumes that we'll use for holding images and working
 // root filesystems for building images.
 func setupContainersStorage(pod *corev1.Pod, container *corev1.Container) {
-	exists := false
+	rootExists, runExists := false, false
 	for _, v := range pod.Spec.Volumes {
 		if v.Name == "container-storage-root" {
-			exists = true
-			break
+			rootExists = true
+		}
+		if v.Name == "container-storage-run" {
+			runExists = true
 		}
 	}
-	if !exists {
+	if !rootExists {
 		pod.Spec.Volumes = append(pod.Spec.Volumes,
 			corev1.Volume{
 				Name: "container-storage-root",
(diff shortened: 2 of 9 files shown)


nalind:
/retitle Run builds with "privileged: false" if BUILD_PRIVILEGED=false

nalind:
/retitle Run builds in user namespaces if BUILD_PRIVILEGED=false

adambkaplan:
/retitle BUILD-433: Run builds in user namespaces if BUILD_PRIVILEGED=false

pkg/build/controller/strategy/custom.go
```diff
@@ -88,7 +88,8 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 		serviceAccount = buildutil.BuilderServiceAccountName
 	}
 
-	privileged := true
+	securityContext := securityContextForBuild(strategy.Env)
+	secretsMode := int32(0600) // todo: change to corev1.SecretVolumeSourceDefaultMode (0644)?
```
adambkaplan, line 92: Do we want to do this for `Custom` strategy builds? Outside of OSBS this isn't really used (though this may be a vote in favor of enabling this capability).

pkg/build/controller/strategy/util.go
```diff
@@ -113,20 +113,20 @@ func setupDockerSocket(pod *corev1.Pod) {
 
 // mountConfigMapVolume is a helper method responsible for actual mounting configMap
 // volumes into a pod.
-func mountConfigMapVolume(pod *corev1.Pod, container *corev1.Container, configMapName, mountPath, volumeSuffix string, volumeSource *corev1.VolumeSource) {
-	mountVolume(pod, container, configMapName, mountPath, volumeSuffix, policy.ConfigMap, volumeSource)
+func mountConfigMapVolume(pod *corev1.Pod, container *corev1.Container, configMapName, mountPath, volumeSuffix string, volumeSource *corev1.VolumeSource, configMapMode int32) {
+	mountVolume(pod, container, configMapName, mountPath, volumeSuffix, policy.ConfigMap, volumeSource, configMapMode)
```
adambkaplan, line 117: We can be opinionated here and mount the ConfigMap with the default volume mode. ConfigMap volumes must be read-only in the container.

pkg/build/controller/strategy/util.go
```diff
@@ -113,20 +113,20 @@ func setupDockerSocket(pod *corev1.Pod) {
...
 }
 
 // mountSecretVolume is a helper method responsible for actual mounting secret
 // volumes into a pod.
-func mountSecretVolume(pod *corev1.Pod, container *corev1.Container, secretName, mountPath, volumeSuffix string, volumeSource *corev1.VolumeSource) {
-	mountVolume(pod, container, secretName, mountPath, volumeSuffix, policy.Secret, volumeSource)
+func mountSecretVolume(pod *corev1.Pod, container *corev1.Container, secretName, mountPath, volumeSuffix string, volumeSource *corev1.VolumeSource, secretMode int32) {
+	mountVolume(pod, container, secretName, mountPath, volumeSuffix, policy.Secret, volumeSource, secretMode)
```
adambkaplan, line 123: Ditto - we can be opinionated and mount with the default mode for Secrets

pkg/build/controller/strategy/util.go
```diff
@@ -113,20 +113,20 @@ func setupDockerSocket(pod *corev1.Pod) {
...
 }
 
 // mountCSIVolume is a helper method responsible for actual mounting csi
 // volumes into a pod.
-func mountCSIVolume(pod *corev1.Pod, container *corev1.Container, volumeName, mountPath, volumeSuffix string, volumeSource *corev1.VolumeSource) {
-	mountVolume(pod, container, volumeName, mountPath, volumeSuffix, policy.CSI, volumeSource)
+func mountCSIVolume(pod *corev1.Pod, container *corev1.Container, volumeName, mountPath, volumeSuffix string, volumeSource *corev1.VolumeSource, csiVolumeMode int32) {
+	mountVolume(pod, container, volumeName, mountPath, volumeSuffix, policy.CSI, volumeSource, csiVolumeMode)
```
adambkaplan, line 129: CSI volumes do not have an associated mode, pass 0 here.

pkg/build/controller/strategy/util.go
```diff
@@ -148,7 +148,6 @@ func mountVolume(pod *corev1.Pod, container *corev1.Container, objName, mountPat
 			break
 		}
 	}
-	mode := int32(0600)
```
adambkaplan, line 151: We should probably change this to the default of 0644 and leave it at that.

@bparees I found this commit set the mode to 0600 - do you recall why?
https://github.com/openshift/openshift-controller-manager/commit/b4cee01a40fdba76d3aa90dbb773da6ccf433258

adambkaplan (commented):
My first pass - I am a bit concerned regarding the refactoring around volume modes - this should be addressed in a separate PR. Is it possible to revert these changes?

pkg/build/controller/strategy/util.go
bparees: i don't recall why we did it specifically, but logically it minimizes who can read the mounted secret and we know there's no reason for any additional `group` or `other` permissions to be granted.  Given these are secrets we're mounting, keeping them as secure as possible seems like the right permission setup.

(it doesn't matter *that* much since the only process running in the pod is our process which has access to the content, but hypothetically it removes some additional exposure avenues)

bparees: i'll also note that it looks like this commit actually went in back in august 2017 (the github ui makes it look like it was 2019).  So it's been this way for a looooong time.

pkg/build/controller/strategy/util.go
nalind: Ah, good to know, thanks!

pkg/build/controller/strategy/util.go
bparees: full history on this can be seen here:
https://github.com/openshift/origin/pull/15248

tldr it wasn't a change from 644 to 0600, it was net new code that replaced existing code that used 0600.  The commit was part of a larger PR.

nalind:
Removed the `error` that `setupBuilderAutonsUser()` could previously return that was always `nil`.

adambkaplan (approved):
/approve

adambkaplan:
/cc @rolfedh 

For docs approved - draft https://github.com/openshift/openshift-docs/pull/44000

adambkaplan:
/cc @jitendar-singh @prietyc123 

For QE approval - should be tested with https://github.com/openshift/builder/pull/291

adambkaplan:
/cc @RickJWagner 

For PX approval

pkg/build/controller/strategy/util.go
```diff
@@ -536,6 +518,50 @@ func addVolumeMountToContainers(conts []corev1.Container, mount corev1.VolumeMou
...
+	securityContext := &corev1.SecurityContext{
+		Privileged: &privileged,
+	}
+	if !privileged {
+		securityContext.Capabilities = &corev1.Capabilities{
+			Drop: []corev1.Capability{
+				"CAP_KILL",
+				"CAP_MKNOD",
+			},
+		}
```
adambkaplan, line 544: Do we need to run as uid/gid 0?
```suggestion
		securityContext.Capabilities = &corev1.Capabilities{
			Drop: []corev1.Capability{
				"CAP_KILL",
				"CAP_MKNOD",
			},
		}
		securityContext.RunAsUser = 0
		securityContext.RunAsGroup = 0
```

nalind: That's the default for the struct, but you're right that there's no harm in making it explicit in the code.  Since that's what we're requesting whether or not we're privileged, should I just set these fields outside of the `if !privileged` block?

nalind: Oh, those are pointers.  Got it.

RickJWagner (approved):
No suggested changes, beyond what's already under discussion.

nalind:
> My 4.11 node only has entries for "core" in those files. That's confusing, because I had assumed that https://bugzilla.redhat.com/show_bug.cgi?id=2038827 was fixed for 4.11 when it was fixed in 4.10, and the "containers" entries are indeed there when I examine a 4.10.3 node running on CRC 2.0.1.

@haircommander's adding them to 4.11 in cri-o-1.24.0-9.rhaos4.11.git77718dc.el8.

rolfedh (approved):
/lgtm from a docs standpoint

nalind:
* Have reverted to making `securityContext.RunAsUser` and `securityContext.RunAsGroup` implicit for unprivileged builds - if they're set explicitly, the service account token's permissions become 0600 instead of 0644, and the builder can't read them.
* Setting up a user namespace with the current set of UIDs/GIDs mapped to themselves (or more specifically, UID 0) to get `CAP_SYS_ADMIN` over their own new mount namespaces requires `CAP_SETFCAP` starting in upstream kernel 5.12 and RHEL kernel 4.18.0-312.el8.  Added it to the unprivileged build security context.

nalind:
* When the pull secret is owned by root with mode `0o600`, an unprivileged builder can't read it.  Tweaked `pkg/build/controller/strategy.mountVolume()` to use mode `0o644` when the security context isn't privileged.

adambkaplan:
@coreydaley if you'd like, I can set up this repo for "no-feature-freeze" so you don't need a BZ to merge.

coreydaley:
@adambkaplan That would be great, thanks!

PR #176, opened 2021-04-20
Bug 1935165: fixed LANG for the builder container

alicerum:
Image on which `builder` is based, has been stripped from a lot of unnecessary files, including some locale files. Because of this, locale `en_US.utf8` does not exist in the builder image, which leads to `bsdtar` not working correctly with non-ansi files in the archive.
`C.utf8` does the job, though, and allows correct work with all the files.

Another possible solution would be installing `glibc-langpack-en` package in the `builder` image, but it would make the image bigger, and it doesn't seem like we really need `en_US` locale.


adambkaplan (approved):
/lgtm

Can confirm on UBI8 that C.utf8 is the only UTF8 locale available.

wewang58:
Launched a cluster to test it, build complete when filename contains an umlaut 
/lgtm

PR #178, opened 2021-04-28
Bug 1954715: Increase number of image-import controller workers

dmage:
When the controller imports an image from a slow registry, it waits for
a response from the registry and almost doesn't consume resources. As
today we have only 5 workers, 5 requests to import from a slow registry
will block the entire queue from beging processed.

### DIFF
--- a/pkg/cmd/controller/image.go
+++ b/pkg/cmd/controller/image.go
@@ -166,7 +166,7 @@ func RunImageImportController(ctx *ControllerContext) (bool, error) {
 		ctx.ClientBuilder.OpenshiftImageClientOrDie(infraImageImportControllerServiceAccountName),
 		informer,
 	)
-	go controller.Run(5, ctx.Stop)
+	go controller.Run(50, ctx.Stop)
 
 	// TODO control this using enabled and disabled controllers
 	if ctx.OpenshiftControllerConfig.ImageImport.DisableScheduledImport {
--- a/pkg/image/controller/factory.go
+++ b/pkg/image/controller/factory.go
@@ -63,7 +63,7 @@ func (opts ScheduledImageStreamControllerOptions) GetRateLimiter() flowcontrol.R
 // NewImageStreamController returns a new image stream import controller.
 func NewImageStreamController(client imagev1client.Interface, informer imagev1informer.ImageStreamInformer) *ImageStreamController {
 	controller := &ImageStreamController{
-		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ImageStreamController"),
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultItemBasedRateLimiter(), "ImageStreamController"),
 
 		client:       client.ImageV1(),
 		lister:       informer.Lister(),


pkg/cmd/controller/image.go
```diff
@@ -166,7 +166,7 @@ func RunImageImportController(ctx *ControllerContext) (bool, error) {
 		ctx.ClientBuilder.OpenshiftImageClientOrDie(infraImageImportControllerServiceAccountName),
 		informer,
 	)
-	go controller.Run(5, ctx.Stop)
+	go controller.Run(100, ctx.Stop)
```
smarterclayton, line 169: 50?  Determined from pending queue?

dmage: 100 is just a random round number, I can put 50, if you want. We also can spawn a go-routine for every incoming event from API server and have only one reader from the queue. It depends on the purpose of this limit, what does it protect?

smarterclayton: It protects us from blowing up from one very bad image stream - i would in general prefer to make progress in pathological cases vs degrade.  timeouts in client code are ok as bounds, but the primary purpose of this queue is to ensure *eventually* something is imported.  not every import has a hard SLA of 15m for example.

pkg/image/controller/factory.go
```diff
@@ -63,7 +63,7 @@ func (opts ScheduledImageStreamControllerOptions) GetRateLimiter() flowcontrol.R
 // NewImageStreamController returns a new image stream import controller.
 func NewImageStreamController(client imagev1client.Interface, informer imagev1informer.ImageStreamInformer) *ImageStreamController {
 	controller := &ImageStreamController{
-		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "ImageStreamController"),
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultItemBasedRateLimiter(), "ImageStreamController"),
```
smarterclayton, line 66: I think this is the right change - I expected us to rate limit streams or namespaces, not globally.

smarterclayton: Is there a way to split workqueue by namespace effectively vs by item?

xiuwang:
In a nightly cluster(which is not include this pr),  creating 121 imagestreamtags, All of them are imported successfully in 2 mins.
@dmage How can I reproduce the scenario? Could you give me an example? Thanks

dmage:
@xiuwang imitate our CI workload: create about 100 imagestreams with ~250 images each (that use `from: DockerImage` in their spec).

xiuwang:
Could reproduce the scenario on old 4.8 version. 1 imagestream with 289 images From., the whole imagestream importing will spend 17 mins.Not to mention creating 100 imagestreams with 250+ image each.

On pr cluster, same imagestream with 289 images, the whole importing spends about 3 mins.

/label qe-approved

pkg/cmd/controller/image.go
```diff
@@ -166,7 +166,7 @@ func RunImageImportController(ctx *ControllerContext) (bool, error) {
 		ctx.ClientBuilder.OpenshiftImageClientOrDie(infraImageImportControllerServiceAccountName),
 		informer,
 	)
-	go controller.Run(5, ctx.Stop)
+	go controller.Run(50, ctx.Stop)
```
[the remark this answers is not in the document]
bparees: nm, i guess if all 5 workers get stuck on a bad registry, that's independent of rate limiting.

PR #185, opened 2021-05-12
WIP - Bug 1897075: Overwriting previous value for a key

ricardomaraschini:
Instead of deleting the old key and schedule the key again in what may be a different bucket we now just overwrite the value in the same bucket.

Prior to this patch, in some situations, a key could jump from one bucket to another before, causing us to never process the event for the key.

This snippet below is a capture of openshift-controller manager, there we can see that the event was scheduled to bucket 2 and when it was about to be processed it got rescheduled to bucket 1:

```
I0512 08:42:57.546007       1 scheduler.go:123] Image controller scheduler: add openshift/cli-artifacts to bucket 2
I0512 08:42:57.639322       1 scheduler.go:79] Image controller scheduler: queue (0):                                                                                                                                
I0512 08:42:57.639442       1 scheduler.go:90] Image controller scheduler: bucket was empty, advance to 1 of 5 buckets
I0512 08:42:57.639450       1 scheduler.go:61] Image controller scheduler: waiting for limit                                                                                                                         
I0512 08:42:57.639468       1 scheduler.go:79] Image controller scheduler: queue (1):       
I0512 08:42:57.639504       1 scheduler.go:90] Image controller scheduler: bucket was empty, advance to 2 of 5 buckets
I0512 08:42:57.639511       1 scheduler.go:61] Image controller scheduler: waiting for limit
I0512 08:44:16.499153       1 scheduler.go:123] Image controller scheduler: add openshift/cli-artifacts to bucket 1
```

### DIFF
--- a/pkg/image/controller/scheduler.go
+++ b/pkg/image/controller/scheduler.go
@@ -91,17 +91,21 @@ func (s *scheduler) next() (interface{}, interface{}, bool) {
 	return nil, nil, true
 }
 
-// Add places the key in the bucket with the least entries (except the current bucket). The key is used to
-// determine uniqueness, while value can be used to associate additional data for later retrieval. An Add
-// removes the previous key and value and will place the item in a new bucket. This allows callers to ensure
-// that Add'ing a new item to the queue purges old versions of the item, while Remove can be conditional on
-// removing only the known old version.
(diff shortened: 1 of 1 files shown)


ricardomaraschini:
We are going to use a different approach to solve this issue. Closing this one.

/close

PR #186, opened 2021-05-13
Updated Dockerfile's registry url and go version

ricardomaraschini:
I have not been able to build this project using the default `Dockerfile` so I am proposing these changes:

- Use images from `registry.ci.openshift.org`
- Updates to use go 1.15 as used in `Dockerfile.rhel`

Please evaluate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.14 AS builder
+FROM registry.ci.openshift.org/ocp/builder:golang-1.15 AS builder
 WORKDIR /go/src/github.com/openshift/openshift-controller-manager
 COPY . .
 RUN make build --warn-undefined-variables
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM registry.ci.openshift.org/openshift/origin-v4.0:base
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.14 AS builder
+FROM registry.ci.openshift.org/ocp/builder:golang-1.15 AS builder
```
gabemontero, line 1: Might make sense to do 

registry.ci.openshift.org/origin/builder:golang-1.15

That requires less authentication than registry.ci.openshift.org/ocp/builder:golang-1.15 and is consisent with 
registry.ci.openshift.org/openshift/origin-v4.0:base

Where as the Dockerfile.rhel file uses the `ocp` namespace, and that requires authentication credentials.

Note, the Dockerfile.rhel is already doing 1.16 ... https://github.com/openshift/openshift-controller-manager/blob/master/Dockerfile.rhel#L1 but when I tried to pull

registry.ci.openshift.org/origin/builder:golang-1.16

it does not exist yet.

dmage: It seems `Dockerfile` can be deleted. Both openshift/release and ocp-build-data use `Dockerfile.rhel`.

ricardomaraschini:
Seems like Dockerfile is not in use.

/close

PR #169, opened 2021-02-12
Bug 1903519: When creating a wildcard ingress the route is never created due to a mismatch in the host format in both resources.…

miheer:
- When creating a wildcard ingress the route is never created due to a mismatch in the host format in both resources.…
- In k8s ingress the hostname is configured as '*.domain.tld' (see [1]) but that's not a valid hostname for a route, and the conversion code copies it literally, failing in the creation.
- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1903519

[1] https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards

PR #187, opened 2021-05-27
route.openshift.io/termination: passthrough  annotation in Ingress not creating a passthrough route

miheer:
route.openshift.io/termination: passthrough  annotation in Ingress not creating a passthrough route

https://bugzilla.redhat.com/show_bug.cgi?id=1878685

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -423,6 +423,14 @@ func (c *Controller) sync(key queueKey) error {
 			continue
 		}
 		for _, path := range rule.HTTP.Paths {
+			pathValue := path.Path
+			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) && len(path.Path) != 0 {
+				//Ignore the ingress and don't create a route
+				continue
+			}
+			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) {
+				pathValue = ""
+			}
 			if path.Backend.Service == nil {
 				// Non-Service backends are not implemented.
 				continue
@@ -436,20 +444,20 @@ func (c *Controller) sync(key queueKey) error {
 			}
 
 			var existing *routev1.Route
-			old, existing = splitForPathAndHost(old, rule.Host, path.Path)
+			old, existing = splitForPathAndHost(old, rule.Host, pathValue)
 			if existing == nil {
-				if r := newRouteForIngress(ingress, &rule, &path, c.secretLister, c.serviceLister); r != nil {
+				if r := newRouteForIngress(ingress, &rule, &path, pathValue, c.secretLister, c.serviceLister); r != nil {
 					creates = append(creates, r)
 				}
 				continue
 			}
 
-			if routeMatchesIngress(existing, ingress, &rule, &path, c.secretLister, c.serviceLister) {
+			if routeMatchesIngress(existing, ingress, &rule, &path, pathValue, c.secretLister, c.serviceLister) {
 				matches = append(matches, existing)
 				continue
 			}
 
-			if r := newRouteForIngress(ingress, &rule, &path, c.secretLister, c.serviceLister); r != nil {
+			if r := newRouteForIngress(ingress, &rule, &path, pathValue, c.secretLister, c.serviceLister); r != nil {
 				// merge the relevant spec pieces
 				preserveRouteAttributesFromExisting(r, existing)
 				updates = append(updates, r)
@@ -574,6 +582,7 @@ func newRouteForIngress(
 	ingress *networkingv1.Ingress,
 	rule *networkingv1.IngressRule,
 	path *networkingv1.HTTPIngressPath,
+	pathValue string,
 	secretLister corelisters.SecretLister,
 	serviceLister corelisters.ServiceLister,
 ) *routev1.Route {
@@ -606,7 +615,7 @@ func newRouteForIngress(
 		},
 		Spec: routev1.RouteSpec{
 			Host: rule.Host,
-			Path: path.Path,
+			Path: pathValue,
 			To: routev1.RouteTargetReference{
 				Name: path.Backend.Service.Name,
 			},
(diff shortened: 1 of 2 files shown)


pkg/route/ingress/ingress.go
```diff
@@ -423,6 +423,9 @@ func (c *Controller) sync(key queueKey) error {
 			continue
 		}
 		for _, path := range rule.HTTP.Paths {
+			if ingress.Annotations["route.openshift.io/termination"] == "passthrough" {
+				path.Path = ""
```
Miciah, line 427: Don't mutate the ingress; define and use new variable `path`, similar to what you did here:  https://github.com/openshift/openshift-controller-manager/pull/169/files#diff-5100b0e98310cb270f5b9b2bc09313e46cdb9c509eef0ea3a76dda2a07ce987bR426-R431

pkg/route/ingress/ingress.go
```diff
@@ -576,6 +579,7 @@ func newRouteForIngress(
 	path *networkingv1.HTTPIngressPath,
 	secretLister corelisters.SecretLister,
 	serviceLister corelisters.ServiceLister,
+	//defaultbackend *networkingv1.IngressBackend
```
Miciah, line 582: This can be deleted.

Miciah (commented):
Is there a unit test to verify that the ingress-to-route controller ignores ingresses that specify passthrough and nonempty paths (where "nonempty" means "not `/`")?

pkg/route/ingress/ingress.go
```diff
@@ -423,6 +423,10 @@ func (c *Controller) sync(key queueKey) error {
 			continue
 		}
 		for _, path := range rule.HTTP.Paths {
+			pathValue := path.Path
+			if ingress.Annotations["route.openshift.io/termination"] == "passthrough" && path.Path == "/" {
```
Miciah, line 427: Use `terminationPolicyAnnotationKey` and `routev1.TLSTerminationPassthrough`:
```suggestion
			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) && path.Path == "/" {
```
I suggested before that we should ignore ingresses that specify a nonempty path (for values of "nonempty" that include `/`) and passthrough termination.  How does the logic you have here handle such ingresses?

pkg/route/ingress/ingress_test.go
```diff
@@ -1094,6 +1094,84 @@ func TestController_sync(t *testing.T) {
...
+			fields: fields{
+				i: &ingressLister{Items: []*networkingv1.Ingress{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:        "1",
+							Namespace:   "test",
+							Annotations: map[string]string{"route.openshift.io/termination": "passthrough"},
+						},
+						Spec: networkingv1.IngressSpec{
+							IngressClassName: &openshiftCustomIngressClassName,
```
Miciah, line 1108: Is there a reason you don't use `openshiftDefaultIngressClassName` here and omit `ic` below?

pkg/route/ingress/ingress_test.go
```diff
@@ -1751,7 +1829,7 @@ func TestController_sync(t *testing.T) {
 						},
 						Spec: routev1.RouteSpec{
 							Host: "test.com",
-							Path: "/",
+							Path: "",
```
Miciah, line 1832: This doesn't look right.  Won't the edge route have `spec.path: /`?

pkg/route/ingress/ingress_test.go
```diff
@@ -3064,9 +3142,9 @@ func TestController_sync(t *testing.T) {
...
```
Miciah, line 3145: Same comment—this change looks wrong.

pkg/route/ingress/ingress.go
```diff
@@ -423,6 +423,14 @@ func (c *Controller) sync(key queueKey) error {
...
 		}
 		for _, path := range rule.HTTP.Paths {
+			pathValue := path.Path
+			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) && len(path.Path) != 0 {
+				//Ignore the ingress and don't create a route
+				continue
+			}
+			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) {
+				pathValue = ""
+			}
```
Miciah, line 433: This can be simplified as follows:
```suggestion
			pathValue := path.Path
			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) {
				if len(path.Path) != 0 {
					//Ignore the ingress and don't create a route
					continue
				}
				pathValue = ""
			}
```
However, I think the following is more correct:
```suggestion
			pathValue := path.Path
			if ingress.Annotations[terminationPolicyAnnotationKey] == string(routev1.TLSTerminationPassthrough) {
				if path.Path != "/" {
					// Passthrough is incompatible with paths.
					continue
				}
				pathValue = ""
			}
```
That is, if the ingress has termination type `passthrough`, then skip it if the path isn't "/" and set the path to the empty string if it is "/".

pkg/route/ingress/ingress_test.go
```diff
@@ -3086,7 +3144,84 @@ func TestController_sync(t *testing.T) {
...
+							Port: &routev1.RoutePort{
+								TargetPort: intstr.FromString("http"),
+							},
+							WildcardPolicy: routev1.WildcardPolicyNone,
+						},
+					},
+				}},
+			},
+			args: queueKey{namespace: "test", name: "1"},
+			wantRouteDeletes: []clientgotesting.DeleteActionImpl{
```
Miciah, line 3222: Why should the route be deleted in this test case?  Shouldn't the edge-terminated route be updated to be a passthrough route?

Miciah (commented):
When you change a test case, please explain why.  

I think I understand why you have been changing the test cases:  If an ingress used edge termination and is updated to use passthrough, then the ingress has an old route with path "/", and the controller looks for a route with path "" (i.e., the empty string), so the new route doesn't match the old route, and so the controller deletes the old route and creates a new route instead of updating the existing route.  Is that the reason you have been changing the test cases?  

Perhaps we need to define a `pathsMatch` function as follows:
```go
// pathsMatch compares two paths and returns a Boolean value indicating whether
// the paths match.  For the purposes of the comparison, "/" and the empty
// string are treated as equal.
func pathsMatch(a, b string) bool {
	if a == "/" {
		a = ""
	}
	if b == "/" {
		b = ""
	}
	return a == b
}
```
Then use `pathsMatch(route.Spec.Path, path)` for the paths comparison in `splitForPathAndHost`, and use `pathsMatch(route.Spec.Path, pathValue)` for the paths comparison in `routeMatchesIngress`.  Does that make sense?

PR #197, opened 2021-09-01
Bug 1977414: fail Build if ConfigMap or Secret does not exist

alicerum:
In case `ConfigMap` or `Secret` in the `Build` `Spec` `Source` does not exist, build will now fail with corresponding reason and message.

There does not seem to be any reason for the build to wait for those resources, as their necessity is known before build creation.

Additionally, in case of the binary input, `Build` will fail if output does not exist.

### DIFF
--- a/pkg/build/controller/build/build_controller.go
+++ b/pkg/build/controller/build/build_controller.go
@@ -160,6 +160,7 @@ type BuildController struct {
 
 	buildStore                      buildv1lister.BuildLister
 	secretStore                     v1lister.SecretLister
+	configMapStore                  v1lister.ConfigMapLister
 	serviceAccountStore             v1lister.ServiceAccountLister
 	podStore                        v1lister.PodLister
 	imageStreamStore                imagev1lister.ImageStreamLister
@@ -207,6 +208,7 @@ type BuildControllerParams struct {
 	ImageStreamInformer                imagev1informer.ImageStreamInformer
 	PodInformer                        kubeinformers.PodInformer
 	SecretInformer                     kubeinformers.SecretInformer
+	ConfigMapInformer                  kubeinformers.ConfigMapInformer
 	ServiceAccountInformer             kubeinformers.ServiceAccountInformer
 	OpenshiftConfigConfigMapInformer   kubeinformers.ConfigMapInformer
 	ControllerManagerConfigMapInformer kubeinformers.ConfigMapInformer
@@ -239,6 +241,7 @@ func NewBuildController(params *BuildControllerParams) *BuildController {
 		imageContentSourcePolicyLister:   params.ImageContentSourcePolicyInformer.Lister(),
 		imageConfigLister:                params.ImageConfigInformer.Lister(),
 		secretStore:                      params.SecretInformer.Lister(),
+		configMapStore:                   params.ConfigMapInformer.Lister(),
 		serviceAccountStore:              params.ServiceAccountInformer.Lister(),
 		podClient:                        params.KubeClient.CoreV1(),
 		configMapClient:                  params.KubeClient.CoreV1(),
@@ -1055,6 +1058,13 @@ func (bc *BuildController) resolveImageReferences(build *buildv1.Build, update *
 		// If we cannot resolve the output reference, the output image stream
 		// may not yet exist. The build should remain in the new state and show the
 		// reason that it is still in the new state.
+
+		// However, if source type is binary, it means, that user is creating it via
+		// the 'oc start-build'. in that case it makes sense to fail the build to let
+		// the user know early that resource does not exist
+		if build.Spec.Source.Type == buildv1.BuildSourceBinary {
+			update.setPhase(buildv1.BuildPhaseFailed)
+		}
 		update.setReason(buildv1.StatusReasonInvalidOutputReference)
 		update.setMessage("Output image could not be resolved.")
 		if err == errNoIntegratedRegistry {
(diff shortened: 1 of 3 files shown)


pkg/build/controller/build/build_controller.go
```diff
@@ -1055,6 +1055,13 @@ func (bc *BuildController) resolveImageReferences(build *buildv1.Build, update *
 		// If we cannot resolve the output reference, the output image stream
 		// may not yet exist. The build should remain in the new state and show the
 		// reason that it is still in the new state.
+
+		// However, if source type is binary, it means, that user is creating it via
+		// the 'oc start-build'. in that case it makes sense to fail the build to let
+		// the user know early that resource does not exist
+		if build.Spec.Source.Type == buildv1.BuildSourceBinary {
+			update.setPhase(buildv1.BuildPhaseFailed)
+		}
```
adambkaplan, line 1064: Two items:

- I would use transitionToPhase here, and move this after the "No integrated registry check" below.
- Does transitioning to "failed" cause `oc start-build` to abort the source upload?

pkg/build/controller/build/build_controller.go
```diff
@@ -2334,6 +2351,53 @@ func (bc *BuildController) configMapDeleted(obj interface{}) {
 	bc.controllerConfigChanged()
 }
 
+// checkConfigMapExists checks whether all ConfigMaps specified by the Build are existing
+// if they are not, return buildUpdate for the failed build and error that led to this
+func (bc *BuildController) checkConfigMapExists(build *buildv1.Build) (*buildUpdate, error) {
+	for _, cm := range build.Spec.Source.ConfigMaps {
+		name := cm.ConfigMap.Name
+		_, err := bc.configMapClient.ConfigMaps(build.Namespace).Get(context.TODO(), name, metav1.GetOptions{})
```
adambkaplan, line 2359: We should add a lister for the ConfigMaps to the build controller and use it here instead of a client. The build controller has a similar lister for Secrets, which comes from an informer that watches all kube objects on the cluster.

pkg/build/controller/build/build_controller.go
```diff
@@ -2334,6 +2351,53 @@ func (bc *BuildController) configMapDeleted(obj interface{}) {
...
+
+		if err != nil {
+			return bc.resourceMissingBuildUpdate("ConfigMap", name, err), err
```
adambkaplan, line 2362: We probably shouldn't return early here. Instead we should collect a list of all missing ConfigMaps (and secrets, too)

pkg/build/controller/build/build_controller.go
```diff
@@ -2334,6 +2351,53 @@ func (bc *BuildController) configMapDeleted(obj interface{}) {
...
+		}
+	}
+
+	return nil, nil
+}
+
+// checkSecretExists checks whether all Secrets specified by the Build are existing
+// if they are not, return buildUpdate for the failed build and error that led to this
+func (bc *BuildController) checkSecretExists(build *buildv1.Build) (*buildUpdate, error) {
```
adambkaplan, line 2371:
```suggestion
// checkSecretsExist checks whether all Secrets specified by the Build are existing
// if they are not, return buildUpdate for the failed build and error that led to this
func (bc *BuildController) checkSecretsExist(build *buildv1.Build) (*buildUpdate, error) {
```

pkg/build/controller/build/build_controller.go
```diff
@@ -2334,6 +2351,53 @@ func (bc *BuildController) configMapDeleted(obj interface{}) {
...
+	for _, cm := range build.Spec.Source.Secrets {
```
adambkaplan, line 2372: Also need to do volumes.

pkg/build/controller/build/build_controller.go
```diff
@@ -2334,6 +2351,53 @@ func (bc *BuildController) configMapDeleted(obj interface{}) {
...
+		name := cm.Secret.Name
+		_, err := bc.secretStore.Secrets(build.Namespace).Get(name)
+
+		if err != nil {
+			return bc.resourceMissingBuildUpdate("Secret", name, err), err
```
adambkaplan, line 2377: Same as above - we should group all the missing resources together.

pkg/build/controller/build/build_controller.go
```diff
@@ -1182,6 +1189,16 @@ func (bc *BuildController) createBuildPod(build *buildv1.Build) (*buildUpdate, e
...
 
+	// Check that ConfigMaps and secrets specified by build Source exist in the project
+	// if they don't exist, fail the build with GenericBuildFailed reason and
+	// corresponding message
+	if update, err := bc.checkConfigMapExists(build); err != nil {
+		return update, err
+	}
+	if update, err := bc.checkSecretExists(build); err != nil {
+		return update, err
+	}
```
adambkaplan, line 1200: The check functions should return a list of missing resources, which then get converted to a BuildUpdate.

adambkaplan (changes requested):
On the right track here.

I think we should use listers instead of clients to check if the resource exists. Listers limit apiserver hits, and we get those for free because other controllers use the listers provided by our cluster-wide kube informer.

pkg/build/controller/build/build_controller.go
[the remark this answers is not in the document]
alicerum: I am not sure about it, because this does not exit the code, it just makes sure build fails in case of `Binary` build and errors regarding output. Comments in this part of code suggest that not failing the build here is intended behavior in order to create `imagestream` or fix the error while build is hanging.

However, maybe it's not a good idea to let the build hang, and just fail build in case of any input type regardless?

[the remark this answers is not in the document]
adambkaplan: Per live discussion - this is fine. `transitionToPhase` doesn't give us a lot of extra benefit, and failing fast here lets the binary build end.

adambkaplan (commented):
Looking at what we're doing with the output imagestream, we should only fail the build if Binary source is being used. A build that uses other source types should not start the build pod in this situation, instead surfacing via the Reason and message that the secets/configMaps are missing.

adambkaplan (approved):
/lgtm

PR #205, opened 2021-10-12
Add metric to count event with reason needPods

martinkennelly:
This metric will count the number of times this
event is seen and give an indication whether
idling is used.

Signed-off-by: Martin Kennelly <mkennell@redhat.com>

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -30,3 +30,8 @@ OCM exposes the following metrics to Prometheus for monitoring and analysis:
 | ---- | ---- | ------ | ----------- |
 | `openshift_template_instance_completed_total` | Counter | `condition` | Counts completed TemplateInstance objects by condition |
 | `openshift_template_instance_active_age_seconds` | Histogram | No labels | Shows the instantaneous age distribution of active TemplateInstance objects |
+
+## Unidling
+| Name | Type | Labels | Description |
+| ---- | ---- | ------ | ----------- |
+| `openshift_unidle_events_total` | Counter | No labels | Total count of unidling events observed by the unidling controller |
--- a/pkg/unidling/controller/unidling_controller.go
+++ b/pkg/unidling/controller/unidling_controller.go
@@ -24,12 +24,14 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/pager"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/component-base/metrics"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/api"
 	unidlingapi "github.com/openshift/api/unidling/v1alpha1"
 	appstypedclient "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	"github.com/openshift/library-go/pkg/unidling/unidlingclient"
+	unidlingmetrics "github.com/openshift/openshift-controller-manager/pkg/unidling/metrics"
 )
 
 const MaxRetries = 5
@@ -73,6 +75,7 @@ type UnidlingController struct {
 	servicesNamespacer  corev1client.ServicesGetter
 	queue               workqueue.RateLimitingInterface
 	lastFiredCache      *lastFiredCache
+	eventsTotal         *metrics.Counter
 
 	// TODO: remove these once we get the scale-source functionality in the scale endpoints
 	dcNamespacer appstypedclient.DeploymentConfigsGetter
@@ -95,6 +98,7 @@ func NewUnidlingController(scaleNS scale.ScalesGetter, mapper meta.RESTMapper, e
 		lastFiredCache: &lastFiredCache{
 			items: make(map[types.NamespacedName]time.Time),
 		},
+		eventsTotal: unidlingmetrics.GetEventsTotalCounter(),
 
 		dcNamespacer: dcNamespacer,
 		rcNamespacer: rcNamespacer,
@@ -212,6 +216,7 @@ func (c *UnidlingController) enqueueEvent(event *corev1.Event) {
 
 	// only add things to the queue if they're newer than what we already have
 	if c.lastFiredCache.AddIfNewer(info, event.LastTimestamp.Time) {
+		c.eventsTotal.Inc()
 		c.queue.Add(info)
 	}
 }
--- a/pkg/unidling/metrics/doc.go
+++ b/pkg/unidling/metrics/doc.go
@@ -1,0 +1,2 @@
+// Package metrics contains code for unidling controller metrics
+package metrics
--- a/pkg/unidling/metrics/metrics.go
+++ b/pkg/unidling/metrics/metrics.go
@@ -1,0 +1,25 @@
+package metrics
+
+import (
+	"sync"
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+var (
+	unidleCount = metrics.NewCounter(&metrics.CounterOpts{
+		Namespace: "openshift",
+		Subsystem: "unidle",
+		Name:      "events_total",
+		Help:      "Total count of unidling events observed by the unidling controller",
+	})
+	registerOnce sync.Once
+)
+
+func GetEventsTotalCounter() *metrics.Counter {
+	registerOnce.Do(func() {
+		legacyregistry.MustRegister(unidleCount)
+	})
+	return unidleCount
+}


martinkennelly:
@danwinship I am looking to publish a representation of how often idling is used. I believe Kubernetes events with a reason of "needPods" is a good representation of how often this feature is used and controller manager is a good place to capture this versus implementing this on the each CNI implementation.

Another option is just to export a metric for when idling is enabled or disabled for each CNI implementation but I think this is a better proxy metric to detect if idling is being used or not and to what rough extent.

You seem to be the expert behind this feature originally. Does this make any sense?

martinkennelly:
Forced pushed to change metric name from ```openshift_unidling_needpods_count``` to ```openshift_unidling_count```

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+	uc = unidlingCollector{}
+)
+
+type unidlingCollector struct {
+	isCreated  bool
+	createOnce sync.Once
+	createLock sync.Mutex
+
+	getUnidlingCountFunc func() uint64
+}
```
adambkaplan, line 29: Won't be needed with component-base's `NewCounter`

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+// Create satisfies the k8s metrics.Registerable interface. It is called when the metric is
+// registered with Prometheus via k8s metrics.
+func (uc *unidlingCollector) Create(v *semver.Version) bool {
+	uc.createOnce.Do(func() {
+		uc.createLock.Lock()
+		defer uc.createLock.Unlock()
+		uc.isCreated = true
+	})
+	return uc.IsCreated()
+}
```
adambkaplan, line 48: Won't be needed with `NewCounter`

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+// Describe implements the prometheus.Collector interface.
+func (uc *unidlingCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- unidlingCountDesc
+}
+
```
adambkaplan, line 54: Won't be needed with `NewCounter`

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+// Collect implements the prometheus.Collector interface.
+func (uc *unidlingCollector) Collect(ch chan<- prometheus.Metric) {
+	if !uc.IsCreated() {
+		return
+	}
+	ch <- prometheus.MustNewConstMetric(unidlingCountDesc, prometheus.CounterValue, float64(uc.getUnidlingCountFunc()))
+}
```
adambkaplan, line 61: Won't be needed with `NewCounter`

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (uc *unidlingCollector) ClearState() {
+	uc.createLock.Lock()
+	defer uc.createLock.Unlock()
+	uc.isCreated = false
+}
```
adambkaplan, line 67: Won't be needed with `NewCounter`

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (uc *unidlingCollector) FQName() string {
+	return "openshift_unidling"
+}
```
adambkaplan, line 71: Won't be needed with `NewCounter`

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+// IsCreated indicates if the unidling metrics were created and registered with Prometheus.
+func (uc *unidlingCollector) IsCreated() bool {
+	return uc.isCreated
+}
```
adambkaplan, line 76: Won't be needed with `NewCounter`

docs/metrics.md
```diff
@@ -30,3 +30,8 @@ OCM exposes the following metrics to Prometheus for monitoring and analysis:
 | ---- | ---- | ------ | ----------- |
 | `openshift_template_instance_completed_total` | Counter | `condition` | Counts completed TemplateInstance objects by condition |
 | `openshift_template_instance_active_age_seconds` | Histogram | No labels | Shows the instantaneous age distribution of active TemplateInstance objects |
+
+## Unidling
+| Name | Type | Labels | Description |
+| ---- | ---- | ------ | ----------- |
+| `openshift_unidling_count` | Counter | No labels | Counts number of events handled with a reason of needPods |
```
adambkaplan, line 37:
```suggestion
| `openshift_unidle_events_total` | Counter | No labels | Total count of unidling events observed by the unidling controller |
```

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+var (
+	unidlingCountDesc = prometheus.NewDesc(
+		"openshift_unidling_count",
+		"The number of times an unidling occurred",
+		nil,
+		nil,
+	)
+	uc = unidlingCollector{}
+)
```
adambkaplan, line 21: The Prometheus SDK has a built-in Counter that is much simpler to use and handles atomic increments. Kubernetes component-base has a wrapper around this which makes it compatible with MustRegister. See https://github.com/kubernetes/component-base/blob/v0.22.3/metrics/counter.go#L41.

Metrics should adhere to Prometheus standards for naming. Counts that do not have units should use the `total` suffix. See https://prometheus.io/docs/practices/naming/.

```suggestion
var (
	unidlingCountMetric = metrics.NewCounter(&CounterOps{
	             Name: "openshift_unidle_events_total",
	             Help: "Total count of unidling events observed by the unidling controller"
	         },
	)
)

func IncrementUnidlingCount() {
    unidlingCountMetric.Inc()
}
```

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+	createLock sync.Mutex
+
+	getUnidlingCountFunc func() uint64
+}
+
+func InitializeMetricsCollector(unidlingCountFunc func() uint64) {
+	if !uc.IsCreated() {
+		uc.getUnidlingCountFunc = unidlingCountFunc
+		legacyregistry.MustRegister(&uc)
+	}
```
adambkaplan, line 35:
```suggestion
	if !unidlingCountMetric.IsCreated() {
		legacyregistry.MustRegister(unidlingCountMetric)
	}
```

pkg/unidling/controller/unidling_controller.go
```diff
@@ -204,6 +205,7 @@ func (c *UnidlingController) enqueueEvent(event *corev1.Event) {
 	if event.Reason != unidlingapi.NeedPodsReason {
 		return
 	}
+	c.needPodsEventCount++
```
adambkaplan, line 208:
```suggestion
	metrics.IncrementUnidlingCount()
```

pkg/unidling/controller/unidling_controller.go
```diff
@@ -73,6 +73,7 @@ type UnidlingController struct {
 	servicesNamespacer  corev1client.ServicesGetter
 	queue               workqueue.RateLimitingInterface
 	lastFiredCache      *lastFiredCache
+	needPodsEventCount  uint64
```
adambkaplan, line 76: Not needed - let Prometheus take care of this.

pkg/unidling/controller/unidling_controller.go
```diff
@@ -438,6 +440,10 @@ func (c *UnidlingController) handleRequest(info types.NamespacedName, lastFired
 	return false, nil
 }
 
+func (c *UnidlingController) GetNeedPodsCount() uint64 {
+	return c.needPodsEventCount
+}
+
```
adambkaplan, line 446: Not needed - let Prometheus take care of this.

pkg/unidling/metrics/metrics.go
```diff
@@ -0,0 +1,76 @@
...
+
+type unidlingCollector struct {
+	isCreated  bool
+	createOnce sync.Once
+	createLock sync.Mutex
+
+	getUnidlingCountFunc func() uint64
+}
+
+func InitializeMetricsCollector(unidlingCountFunc func() uint64) {
```
adambkaplan, line 31: When Promethus handles the counter, no function is needed here.

```suggestion
func InitializeMetricsCollector() {
```

adambkaplan (changes requested):
It seems that we have a bit of tech debt to clean up when it comes to counter metrics in openshift-controller-manager. The Prometheus SDK (wrapped by Kubernetes component-base) has first class support for Counter metrics, which takes care of atomic increments and such. There is no need to implement a separate collector (which we seem to do for the imagestream import metrics).

I have also suggested a metric name which conforms to the Prometheus naming standards - we'll need the monitoring team to sign off on this before merging.

pkg/unidling/metrics/metrics.go
martinkennelly: @adambkaplan Thank you for the review. Appreciated. I originally did not think I could use Prometheus counter with the existing way this component publishes metrics. Thank you so much for dramatically simplifying my PR.

pkg/cmd/controller/unidling.go
```diff
@@ -42,5 +43,6 @@ func RunUnidlingController(ctx *ControllerContext) (bool, error) {
 
 	go controller.Run(ctx.Stop)
 
+	unidlingMetrics.InitializeUnidlingControllerMetrics()
```
adambkaplan, line 46: I think you'll want to initialize the metric before you run the controller in a goroutine. Othwerwise you risk a race condition incrementing the metric.

adambkaplan (commented):
/approve

One comment about when the metric is initialized, otherwise looks good from me.

adambkaplan:
@openshift/openshift-team-monitoring ptal for naming/cardinality review.

simonpasquier (commented):
the metric looks good to me. Out of curiosity, any reason why the counter couldn't a field of the `UnidlingController` struct directly? e.g. `c.unidleCounter.Inc()` instead of `unidlingMetrics.IncrementUnidleCount()`

martinkennelly:
> the metric looks good to me. Out of curiosity, any reason why the counter couldn't a field of the `UnidlingController` struct directly? e.g. `c.unidleCounter.Inc()` instead of `unidlingMetrics.IncrementUnidleCount()`

@simonpasquier No reason why I didn't do that and it makes sense when there is only going to be one consumer of this metric. It would have been a little bit cleaner I think because no need for the pointless wrapper `IncrementUnidleCount` func then. Do you want me to change it? Its a minor nit for sure but I don't mind doing it.

simonpasquier:
@martinkennelly if you can update, it would be great since it's much more the way the Prometheus client library works. Other than that, the metric itself is good to me.

martinkennelly:
@adambkaplan we are ready to go here. Thank you for your time. 
If you want me to update anything in controller manager, let me know. Happy to do it when I have time.

pkg/unidling/controller/unidling_controller.go
```diff
@@ -24,12 +24,14 @@ import (
...
 	"k8s.io/client-go/tools/pager"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/component-base/metrics"
 	"k8s.io/klog/v2"
 
 	"github.com/openshift/api"
 	unidlingapi "github.com/openshift/api/unidling/v1alpha1"
 	appstypedclient "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
 	"github.com/openshift/library-go/pkg/unidling/unidlingclient"
+	unidlingMetrics "github.com/openshift/openshift-controller-manager/pkg/unidling/metrics"
```
adambkaplan, line 34: Style nit - camelCase isn't typically used in imports.

```suggestion
	unidlingmetrics "github.com/openshift/openshift-controller-manager/pkg/unidling/metrics"
```

pkg/unidling/controller/unidling_controller.go
```diff
@@ -95,6 +98,7 @@ func NewUnidlingController(scaleNS scale.ScalesGetter, mapper meta.RESTMapper, e
 		lastFiredCache: &lastFiredCache{
 			items: make(map[types.NamespacedName]time.Time),
 		},
+		eventsTotal: unidlingMetrics.GetEventsTotalCounter(),
```
adambkaplan, line 101: Similar style nit
```suggestion
		eventsTotal: unidlingmetrics.GetEventsTotalCounter(),
```

pkg/unidling/metrics/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package prometheus contains code for unidling related metrics
+// into the prometheus registry running in the openshift master
+package prometheus
```
adambkaplan, line 3: Style and language nits:
- Package name should match the directory name, 
- Avoid the use of "master" as this is not inclusive. "Control plane" is the preferred term for the core components of Kubernetes and OpenShift.

```suggestion
// Package metrics contains code for unidling controller metrics
package metrics
```

pkg/cmd/controller/unidling.go
```diff
@@ -39,6 +40,7 @@ func RunUnidlingController(ctx *ControllerContext) (bool, error) {
 		coreClient,
 		resyncPeriod,
 	)
+	unidlingMetrics.GetEventsTotalCounter()
```
adambkaplan, line 43: Since `NewUnidlingController` calls `GetEventsTotalCounter`, is this necessary?

adambkaplan (changes requested):
@martinkennelly mostly style nits here, plus a question regarding some of the initialization code.

adambkaplan:
/retest

Unrelated flakes.

pkg/cmd/controller/unidling.go
martinkennelly: No - removed.

adambkaplan (approved):
/lgtm

Builds currently failing on an unrelated flake:

```
event happened 22 times, something is wrong: ns/openshift-controller-manager daemonset/controller-manager - reason/SuccessfulDelete (combined from similar events): Deleted pod: controller-manager-74rw5
```

PR #210, opened 2021-11-24
BUILD-275: add support for build csi volume source

jkhelil:
https://issues.redhat.com/browse/BUILD-275


adambkaplan (commented):
@jkhelil generally on the right path here. You can get the build to pass by adding a commit which adds a replace to your `go.mod` file, referencing your code in https://github.com/openshift/api/pull/1056

(note that your code is currently missing the deep copy functions, and will break as a result.)

jkhelil:
flaking openshift-e2e-aws-builds-techpreview 
/test openshift-e2e-aws-builds-techpreview

jkhelil:
flaking e2e-gcp-builds 
/test e2e-gcp-builds

gabemontero (commented):
You should put `BUILD-275: ` in the beginning of your PR title.  I know you reference it in the PR description, but having it in the title will help us when we have to search PRs in the future.

I do have a question for @adambkaplan in looking at the BUILD-275 acceptance criteria.

It mentions making sure that `TechPreviewNoUpgrade` feature gate is enabled.

However, the `csiVolumesEnabled` parameter to `setupBuildVolumes` is currently on based on only the `BuildCSIVolumes` feature gate.

See https://github.com/openshift/openshift-controller-manager/blob/ebe986727319c6c7741cc841bff7fd4f7801c7ca/pkg/cmd/controller/build.go#L54-L63

@adambkaplan - should that code be changed to check both feature gates?  I don't *think* that `BuildCSIVolumes` is automatically set to true if tech preview is on, but maybe I am "misremembering" .... and I don't recall exactly where you wired all that feature gate stuff up.

Otherwise @jkhelil you changes here look OK to me wrt wiring in CSI volumes.  Though I think getting a review from @coreydaley since he did the introduction of build volumes in general would be good, in case I am missing something.

pkg/build/controller/strategy/util.go
```diff
@@ -189,6 +195,12 @@ func makeVolume(volumeName, refName string, mode int32, fsType policy.FSType, vo
 				DefaultMode: &mode,
 			}
 		}
+	case policy.CSI:
+		if volumeSource != nil && volumeSource.CSI != nil {
+			vol.VolumeSource.CSI = volumeSource.CSI.DeepCopy()
```
gabemontero, line 200: depending on where the discussion thread in https://github.com/openshift/origin/pull/26646#discussion_r763126357 lands, we may perhaps want to force the setting of the `refreshResource` attribute to 'false' for any shared resource CSI volumes.

gabemontero: in our team meeting yesterday, we reached the consensus that at this time at least we were not going to force `refreshResources` to be set to false

gabemontero:
> You should put `BUILD-275: ` in the beginning of your PR title. I know you reference it in the PR description, but having it in the title will help us when we have to search PRs in the future.
> 
> I do have a question for @adambkaplan in looking at the BUILD-275 acceptance criteria.
> 
> It mentions making sure that `TechPreviewNoUpgrade` feature gate is enabled.
> 
> However, the `csiVolumesEnabled` parameter to `setupBuildVolumes` is currently on based on only the `BuildCSIVolumes` feature gate.
> 
> See
> 
> https://github.com/openshift/openshift-controller-manager/blob/ebe986727319c6c7741cc841bff7fd4f7801c7ca/pkg/cmd/controller/build.go#L54-L63
> 
> @adambkaplan - should that code be changed to check both feature gates? I don't _think_ that `BuildCSIVolumes` is automatically set to true if tech preview is on, but maybe I am "misremembering" .... and I don't recall exactly where you wired all that feature gate stuff up.
> 

@adambkaplan assisted this old man's memory yesterday (including reminding me of the fact that I actually did some of this stuff, albeit more or less following his instructions ;-) ), and confirmed that only checking for `BuildCSIVolumes` is sufficient for our means.  It s a feature set vs. feature gate distinction, where we can't have BuildCSIVolumes on unless tech preview is on.

So all good on this point.

pkg/build/controller/strategy/util.go
adambkaplan: I concur - the build controller should be agnostic with respect to the specific CSI driver used to provide the volume.

adambkaplan (approved):
/approve

Controller changes look good. Once the code for openshift/api is merged then we can use a proper commit to get the updated Build API object.

adambkaplan:
/hold

Need https://github.com/openshift/api/pull/1056 to merge, as well as an update to openshift/apiserver.

PR #213, opened 2021-12-21
Bug 2034484: fix:library-go nil pointer issue

qJkee:
pull library-go in to grab fix described here https://github.com/openshift/library-go/pull/1272


soltysh (approved):
/lgtm
/approve

PR #216, opened 2022-01-24
Add support for InsecureEdgeTerminationPolicy

holzman:
Set the InsecureEdgeTermationPolicy to None, Allow, Redirect with
the route.openshift.io/insecureEdgeTerminationPolicy annotation


holzman:
Just a proof-of-concept (based on a comment on #openshift-users on the k8s slack workspace).

PR #218, opened 2022-03-11
NE-860: ingress-to-route: add support for destinationCACertificate

lechuk47:
## Overview

This PR is the continuation of #153. It adds the requests made on the original PR.

* The annotation now is a reference to a secret. The secret contains the CA to validate the backend certificate
* Added tests for this feature. 
* I've considered to change the annotation name to include "Secret" at the end. 

## Example 

```yaml
kind: Ingress
metadata:
  annotations:
    route.openshift.io/destination-ca-certificate-secret: <existing-secret-in-the-same-namespace>
...
```

will generate: 

```yaml
kind: Route
metadata:
  annotations:
    route.openshift.io/destination-ca-certificate-secret: <existing-secret-in-the-same-namespace>
spec:
 ...
  tls:
    termination: reencrypt
    destinationCACertificate: |
        `tls.crt key secret contents`
   ...
```
 
let me know if something else is needed.

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -635,10 +635,12 @@ func preserveRouteAttributesFromExisting(r, existing *routev1.Route) {
 	r.Spec.To.Weight = existing.Spec.To.Weight
 	if r.Spec.TLS != nil && existing.Spec.TLS != nil {
 		r.Spec.TLS.CACertificate = existing.Spec.TLS.CACertificate
-		if _, ok := r.Annotations[destinationCACertificateAnnotationKey]; !ok {
-			r.Spec.TLS.DestinationCACertificate = existing.Spec.TLS.DestinationCACertificate
-		}
 		r.Spec.TLS.InsecureEdgeTerminationPolicy = existing.Spec.TLS.InsecureEdgeTerminationPolicy
+		if r.Spec.TLS.Termination == routev1.TLSTerminationReencrypt {
+			if _, ok := r.Annotations[destinationCACertificateAnnotationKey]; !ok {
+				r.Spec.TLS.DestinationCACertificate = existing.Spec.TLS.DestinationCACertificate
+			}
+		}
 	}
 }
 
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -900,7 +900,6 @@ func terminationPolicyForIngress(ingress *networkingv1.Ingress) routev1.TLSTermi
 	}
 }
 
-var destinationCACertificateSecretKeyName = "destinationCACertificate"
 var destinationCACertificateAnnotationKey = routev1.GroupName + "/destination-ca-certificate-secret"
 
 func destinationCACertificateForIngress(ingress *networkingv1.Ingress, secretLister corelisters.SecretLister) *string {
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -899,10 +899,12 @@ func destinationCACertificateForIngress(ingress *networkingv1.Ingress, secretLis
 	if err != nil {
 		return nil
 	}
-	if secret.Type != corev1.SecretTypeOpaque {
+	switch secret.Type {
+	case corev1.SecretTypeTLS, corev1.SecretTypeOpaque:
+	default:
 		return nil
 	}
-	if v, ok := secret.Data[destinationCACertificateSecretKeyName]; ok {
+	if v, ok := secret.Data[corev1.TLSCertKey]; ok {
 		value := string(v)
 		return &value
 	}
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -623,7 +623,7 @@ func newRouteForIngress(
 				Name: path.Backend.Service.Name,
 			},
 			Port:           port,
-			TLS:            tlsConfigForIngress(ingress, rule, tlsSecret),
+			TLS:            tlsConfigForIngress(ingress, rule, tlsSecret, secretLister),
 			WildcardPolicy: wildcardPolicy,
 		},
 	}
@@ -683,7 +683,8 @@ func routeMatchesIngress(
 	if hasInvalidTLSSecret {
 		return false
 	}
-	tlsConfig := tlsConfigForIngress(ingress, rule, tlsSecret)
+
+	tlsConfig := tlsConfigForIngress(ingress, rule, tlsSecret, secretLister)
 	if route.Spec.TLS != nil && tlsConfig != nil {
 		tlsConfig.InsecureEdgeTerminationPolicy = route.Spec.TLS.InsecureEdgeTerminationPolicy
 	}
@@ -810,6 +811,7 @@ func tlsConfigForIngress(
 	ingress *networkingv1.Ingress,
 	rule *networkingv1.IngressRule,
 	potentiallyNilTLSSecret *corev1.Secret,
+	secretLister corelisters.SecretLister,
 ) *routev1.TLSConfig {
 	if !tlsEnabled(ingress, rule, potentiallyNilTLSSecret) {
 		return nil
@@ -826,6 +828,12 @@ func tlsConfigForIngress(
 		tlsConfig.Certificate = string(potentiallyNilTLSSecret.Data[corev1.TLSCertKey])
 		tlsConfig.Key = string(potentiallyNilTLSSecret.Data[corev1.TLSPrivateKeyKey])
 	}
+
+	destinationCACertificate := destinationCACertificateForIngress(ingress, secretLister)
+	if terminationPolicy == routev1.TLSTerminationReencrypt && destinationCACertificate != nil {
+		tlsConfig.DestinationCACertificate = *destinationCACertificate
+	}
+
 	return tlsConfig
 }
 
@@ -871,6 +879,7 @@ func tlsSecretIfValid(ingress *networkingv1.Ingress, rule *networkingv1.IngressR
 }
 
 var terminationPolicyAnnotationKey = routev1.GroupName + "/termination"
+var destinationCACertificateAnnotationKey = routev1.GroupName + "/destinationCACertificate"
 
 func terminationPolicyForIngress(ingress *networkingv1.Ingress) routev1.TLSTerminationType {
 	switch {
@@ -881,4 +890,20 @@ func terminationPolicyForIngress(ingress *networkingv1.Ingress) routev1.TLSTermi
 	default:
 		return routev1.TLSTerminationEdge
 	}
+}
+func destinationCACertificateForIngress(ingress *networkingv1.Ingress, secretLister corelisters.SecretLister) *string {
+	name := ingress.Annotations[destinationCACertificateAnnotationKey]
+	secretDataKey := "destinationCACertificate"
+	secret, err := secretLister.Secrets(ingress.Namespace).Get(name)
+	if err != nil {
+		return nil
+	}
+	if secret.Type != corev1.SecretTypeOpaque {
+		return nil
+	}
+	if v, ok := secret.Data[secretDataKey]; ok {
+		value := string(v)
+		return &value
+	}
+	return nil
 }
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -891,7 +891,7 @@ func terminationPolicyForIngress(ingress *networkingv1.Ingress) routev1.TLSTermi
 }
 
 var destinationCACertificateSecretKeyName = "destinationCACertificate"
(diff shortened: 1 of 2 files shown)


lechuk47:
@Miciah @frobware Hi, Is there any option to get this reviewed?

pkg/route/ingress/ingress.go
```diff
@@ -635,7 +635,6 @@ func preserveRouteAttributesFromExisting(r, existing *routev1.Route) {
 	r.Spec.To.Weight = existing.Spec.To.Weight
 	if r.Spec.TLS != nil && existing.Spec.TLS != nil {
 		r.Spec.TLS.CACertificate = existing.Spec.TLS.CACertificate
-		r.Spec.TLS.DestinationCACertificate = existing.Spec.TLS.DestinationCACertificate
```
Miciah, line 638: We need to preserve this field for routes that don't have the annotation because if we don't, we will break any users who rely on the preservation of this field.  

Please add the following test case to make sure we don't break these users:
```go
		{
			name: "no-op - destination CA certificate has been changed by the user",
			fields: fields{
				i: &ingressLister{Items: []*networkingv1.Ingress{
					{
						ObjectMeta: metav1.ObjectMeta{
							Name:      "1",
							Namespace: "test",
							Annotations: map[string]string{
								"route.openshift.io/termination": "reencrypt",
							},
						},
						Spec: networkingv1.IngressSpec{
							TLS: []networkingv1.IngressTLS{
								{Hosts: []string{"test.com"}, SecretName: "secret-1"},
							},
							Rules: []networkingv1.IngressRule{
								{
									Host: "test.com",
									IngressRuleValue: networkingv1.IngressRuleValue{
										HTTP: &networkingv1.HTTPIngressRuleValue{
											Paths: []networkingv1.HTTPIngressPath{
												{
													Path:     "/",
													PathType: &pathTypePrefix,
													Backend: networkingv1.IngressBackend{
														Service: &networkingv1.IngressServiceBackend{
															Name: "service-1",
															Port: networkingv1.ServiceBackendPort{
																Name: "http",
															},
														},
													},
												},
											},
										},
									},
								},
							},
						},
						Status: networkingv1.IngressStatus{
							LoadBalancer: v1.LoadBalancerStatus{
								Ingress: []v1.LoadBalancerIngress{
									{Hostname: "apps.foo.com"},
								},
							},
						},
					},
				}},
				r: &routeLister{Items: []*routev1.Route{
					{
						ObjectMeta: metav1.ObjectMeta{
							Name:            "1-abcdef",
							Namespace:       "test",
							OwnerReferences: []metav1.OwnerReference{{APIVersion: "networking.k8s.io/v1", Kind: "Ingress", Name: "1", Controller: &boolTrue}},
							Annotations: map[string]string{
								"route.openshift.io/termination": "reencrypt",
							},
						},
						Spec: routev1.RouteSpec{
							Host: "test.com",
							Path: "/",
							To: routev1.RouteTargetReference{
								Name: "service-1",
							},
							Port: &routev1.RoutePort{
								TargetPort: intstr.FromString("http"),
							},
							WildcardPolicy: routev1.WildcardPolicyNone,
							TLS: &routev1.TLSConfig{
								Termination: routev1.TLSTerminationReencrypt,
								Key:                      "key",
								Certificate:              "cert",
								DestinationCACertificate: "cert",
							},
						},
						Status: routev1.RouteStatus{
							Ingress: []routev1.RouteIngress{
								{
									RouterCanonicalHostname: "apps.foo.com",
									Conditions: []routev1.RouteIngressCondition{{
										Type:   routev1.RouteAdmitted,
										Status: v1.ConditionTrue,
									}},
								},
							},
						},
					},
				}},
			},
			args: queueKey{namespace: "test", name: "1"},
		},
```

Then, make sure we still preserve the field in the annotation's absence:

```suggestion
		if _, ok := r.Annotations[destinationCACertificateAnnotationKey]; !ok {
			r.Spec.TLS.DestinationCACertificate = existing.Spec.TLS.DestinationCACertificate
		}
```

In addition, you will need to modify `routeMatchesIngress` likewise:

```diff
 	tlsConfig := tlsConfigForIngress(ingress, rule, tlsSecret, secretLister)
 	if route.Spec.TLS != nil && tlsConfig != nil {
 		tlsConfig.InsecureEdgeTerminationPolicy = route.Spec.TLS.InsecureEdgeTerminationPolicy
+		if _, ok := ingress.Annotations[destinationCACertificateAnnotationKey]; !ok {
+			tlsConfig.DestinationCACertificate = route.Spec.TLS.DestinationCACertificate
+		}
 	}
 	return reflect.DeepEqual(tlsConfig, route.Spec.TLS)
 }
```

pkg/route/ingress/ingress.go
```diff
@@ -882,3 +889,22 @@ func terminationPolicyForIngress(ingress *networkingv1.Ingress) routev1.TLSTermi
 		return routev1.TLSTerminationEdge
 	}
 }
+
+var destinationCACertificateSecretKeyName = "destinationCACertificate"
+var destinationCACertificateAnnotationKey = routev1.GroupName + "/destinationCACertificateSecret"
```
Miciah, line 894: Existing route and ingress annotation keys use dotted.case, snake_case, or kebab-case.  I'd prefer we not add camelCase to the menagerie.  I suggest using kebab-case because it appears to be the predominant style, and it is the style we have used in most recently added annotation keys for routes.  
```suggestion
var destinationCACertificateAnnotationKey = routev1.GroupName + "/destination-ca-certificate-secret"
```

pkg/route/ingress/ingress.go
```diff
@@ -623,7 +623,7 @@ func newRouteForIngress(
 				Name: path.Backend.Service.Name,
 			},
 			Port:           port,
-			TLS:            tlsConfigForIngress(ingress, rule, tlsSecret),
+			TLS:            tlsConfigForIngress(ingress, rule, tlsSecret, secretLister),
```
Miciah, line 626: It seems a little strange to add `secretLister` to `tlsConfigForIngress`'s parameters, with the result that the caller uses `secretLister` to get `tlsSecret` and then passes both `tlsSecret` and `secretLister` to `tlsConfigForIngress`.  Would one of the following refactorings make sense?

* Have the caller get `destinationCACertificate` and pass it instead of `secretLister` to `tlsConfigForIngress`, xor
* have the caller pass `secretLister` to `tlsConfigForIngress`, and then have `tlsConfigForIngress` look up both `tlsSecret` and `destinationCACertificate`.

The latter would require also changing `tlsConfigForIngress` to return `hasInvalidTLSSecret` and would require some minor shuffling, but I imagine it would result in slightly less code duplication in the end.

pkg/route/ingress/ingress.go
```diff
@@ -882,3 +889,22 @@ func terminationPolicyForIngress(ingress *networkingv1.Ingress) routev1.TLSTermi
...
+func destinationCACertificateForIngress(ingress *networkingv1.Ingress, secretLister corelisters.SecretLister) *string {
+	name := ingress.Annotations[destinationCACertificateAnnotationKey]
+	secret, err := secretLister.Secrets(ingress.Namespace).Get(name)
+	if err != nil {
+		return nil
+	}
+	if secret.Type != corev1.SecretTypeOpaque {
+		return nil
+	}
+	if v, ok := secret.Data[destinationCACertificateSecretKeyName]; ok {
```
Miciah, line 905: Is there a reason not to use the more standard `tls.crt` as the data key name?  We should indeed allow using `Opaque`-type secrets so that the user isn't forced to provide both `tls.crt` and `tls.key` when we only need the former.  However, we might as well also allow using `kubernetes.io/tls`-type secrets (though we don't use the `tls.key`), which some users may prefer and already be using to store the destination CA certificate.  

```suggestion
	switch secret.Type {
	case corev1.SecretTypeTLS, corev1.SecretTypeOpaque:
	default:
		return nil
	}
	if v, ok := secret.Data[corev1.TLSCertKey]; ok {
```

Note also that `NewController` sets up a watch on secrets with a filter that excludes `Opaque`-type secrets.  If we want the ingress-to-route controller to update the route when the secret that the ingress references is modified, then we will need to change that.  We should also add a test case for the same, if we decide to implement it.   However, maybe we can leave updating when the referenced secret changes as a follow-up change if it's too complex to do in this PR.

Miciah (commented):
Thanks!  This looks great.  However, we need to be very careful not to break users who rely on preservation of `spec.tls.destinationCACertificate`.  I commented on that and a few more minor issues.  Can you make the suggested changes?

Miciah:
/retitle NE-860: ingress-to-route: add suport for destinationCACertificate

Miciah:
/retitle [NE-860](https://issues.redhat.com/browse/NE-860): ingress-to-route: add support for destinationCACertificate

Miciah:
/retitle NE-860: ingress-to-route: add support for destinationCACertificate

pkg/route/ingress/ingress.go
lechuk47: Added the test case and the suggestion. Github doesn't allow to apply this suggestion directly but is included in the PR.

pkg/route/ingress/ingress.go
lechuk47: I Agree. I applied the second way.

pkg/route/ingress/ingress.go
lechuk47: Changed the kebab-case

pkg/route/ingress/ingress.go
lechuk47: I've added this functionality, now the filter will considers the Opaque type secrets and a Secret type switch could happen (Added a test case too).

gcs278 (commented):
Overall looks good to me, just had some comments on test coverage.

pkg/route/ingress/ingress_test.go
```diff
@@ -888,6 +909,84 @@ func TestController_sync(t *testing.T) {
 				},
 			},
 		},
+		{
```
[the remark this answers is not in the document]
lechuk47: There will be an update to add the annotation but the TLS spec won't be modified. Is this the desired behavior or it's preferable to prevent the annotation sync if the termination type is not reencrypt?

pkg/route/ingress/ingress_test.go
```diff
@@ -1928,6 +2119,179 @@ func TestController_sync(t *testing.T) {
...
+			name: "update route - termination policy changed to reencrypt with destCaCertCertificate",
```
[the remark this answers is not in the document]
lechuk47: I added a test to cover this case. I've modified the `preserveRouteAttributesFromExisting` method to just preserve the `destinationCaCertificate` attribute if the termination is still `reencrypt`.  This collides a bit with Micia's suggestion to preserve this field if it has been patched manually on the route, but I suppose this only makes sense if the termination does not change.

pkg/route/ingress/ingress_test.go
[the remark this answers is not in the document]
gcs278: Hmm I see your updates. I have to think about that, but the reason I brought up is because of your change in `tlsConfigForIngress()`:
```
	destinationCACertificate := destinationCACertificateForIngress(ingress, secretLister)
	if terminationPolicy == routev1.TLSTerminationReencrypt && destinationCACertificate != nil {
		tlsConfig.DestinationCACertificate = *destinationCACertificate
	}
```

I believe translates to: new routes with any type other than type "reencrypt" will not have their TLS spec updated with `DestinationCACertificate`. That's why I brought up the idea of these tests.

I think I agree with your update. It's basically dropping `DestinationCACertificate` from the route if has been updated to be anything other than re-encrypt. I think it's important to be consistent either way. You won't create `DestinationCACertificate` for brand new routes with type passthrough or edge, so it makes sense you would drop it if it switch from reencrypt to passthrough or edge.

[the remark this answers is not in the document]
gcs278: But thinking about @Miciah suggestions, he wants to make sure we preserve `DestinationCACertificate` if the user adds it to the route themselves, I start leaning the other way (don't clear `DestinationCACertificate` if a user defines it and changes the termination type to something other than reencrypt). It might be bad UX for the user to put a `DestinationCACertificate` on the route only for it to disappear.

@Miciah could you weigh in? If we switch the termination to something other than , should we clear the route's `DestinationCACertificate`?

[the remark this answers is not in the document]
Miciah: Good question.  Yeah, if we change the termination type from "reencrypt", then we need to clear `spec.tls.destinationCACertificate` so as not to get a validation error: https://github.com/openshift/openshift-apiserver/blob/db22c038b9184b4ae823bba3f57b669526999a2b/pkg/route/apis/route/validation/validation.go#L253-L260

Miciah:
Thanks!
/approve
/lgtm

PR #219, opened 2022-04-05
Bug 2067868: Update prometheus client_golang from 1.11.0 => 1.11.1

coreydaley:


go.mod
```diff
@@ -67,4 +67,5 @@ replace (
 	k8s.io/mount-utils => k8s.io/mount-utils v0.23.0
 	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.23.0
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.23.0
+	github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.11.1 // CVE-2022-21698
```
adambkaplan, line 70: We have an explicit dependency on client_golang in the `require` section above, please change the dependency there.

adambkaplan (changes requested):
/lgtm cancel

go.mod
coreydaley: updated, ptal

adambkaplan (approved):
/approve

/lgtm

PR #220, opened 2022-04-14
Bug 2075584: Bubble actual error message up to the build

coreydaley:
Include the error string in the message that is set on the build
so that it is easily seen by the user, instead of them having to
dig through the OCM controller logs to find it.


gabemontero:
/lgtm

as @coreydaley mentioned in scrum, the openshift/origin e2e's now have to be tweaked since some condition messages have been tweaked

coreydaley:
/hold
Blocked by https://github.com/openshift/origin/pull/27028

PR #221, opened 2022-04-28
Image trigger handler: Include the object name on error

alvaroaleman:
If there is an error, the handler will currently log something like
```
 unable to extract cache data from *v1.Deployment: unexpected end of JSON inpu
```

which is very difficult to map to an actual object. This PR changes that
error to also include the namespace and name of the object, if possible.


alvaroaleman:
/assign @bparees 
Oleg can not approve it appears

PR #223, opened 2022-05-06
OCPBUGS-5744: dockercfg: use tokenrequest instead of SA secrets

stlaz:
this PR makes use of the TokenRequest API to get a token for the dockercfg SA secret rather than creating a secret and polluting etcd for that very same purpose,


stlaz:
if the secrets are actually being mounted this will not work because the tokens inside the secret need to be rotated

s-urbaniak:
> Possible issue: we need to rotate the token inside the secret, is the change in the secret propagated in the pod?

generally, yes: https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically. But the pod is responsible for re-reading the value, either using a time based approach or by fsnotify.

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -63,14 +65,22 @@ type DockercfgControllerOptions struct {
 	// This is normally signaled from the DockerRegistryServiceController which watches for updates to the internal
 	// container image registry service.
 	DockerURLsInitialized chan struct{}
+
```
mfojtik, line 68: please add godoc, explain this field, so future you will be happy :-)

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -421,107 +462,95 @@ func (e *DockercfgController) createTokenSecret(serviceAccount *v1.ServiceAccoun
...
-	return token, len(token.Data[v1.ServiceAccountTokenKey]) > 0, nil
-}
 
-// createDockerPullSecret creates a dockercfg secret based on the token secret
-func (e *DockercfgController) createDockerPullSecret(serviceAccount *v1.ServiceAccount) (*v1.Secret, bool, error) {
-	tokenSecret, isPopulated, err := e.createTokenSecret(serviceAccount)
+	currentSecret, err = e.client.CoreV1().Secrets(serviceAccount.Namespace).Get(context.TODO(), dockercfgSecret.Name, metav1.GetOptions{})
 	if err != nil {
-		return nil, false, err
+		return nil, false, fmt.Errorf("failed to retrieve the dockercfg secret that wasa supposed to be already created: %w", err)
```
mfojtik, line 522: typo (`was`)

stlaz:
/retest
should be showing better test pass rate as creating tokens should now (master is kube 1.24) be allowed

pkg/serviceaccounts/controllers/deleted_token_secrets.go
```diff
@@ -1,101 +0,0 @@
...
-
-// DockercfgTokenDeletedControllerOptions contains options for the DockercfgTokenDeletedController
-type DockercfgTokenDeletedControllerOptions struct {
-	// Resync is the time.Duration at which to fully re-list secrets.
-	// If zero, re-list will be delayed as long as possible
-	Resync time.Duration
-}
-
-// NewDockercfgTokenDeletedController returns a new *DockercfgTokenDeletedController.
-func NewDockercfgTokenDeletedController(secrets informers.SecretInformer, cl kclientset.Interface, options DockercfgTokenDeletedControllerOptions) *DockercfgTokenDeletedController {
```
deads2k, line 27: Without this controller, what deletes the token secrets when the dockercfg secret is deleted later on?

pkg/serviceaccounts/controllers/docker_registry_service.go
```diff
@@ -138,7 +138,7 @@ type DockerRegistryServiceController struct {
 }
 
 // Runs controller loops and returns immediately
-func (e *DockerRegistryServiceController) Run(workers int, stopCh <-chan struct{}) {
+func (e *DockerRegistryServiceController) Run(ctx context.Context, workers int) {
```
deads2k, line 141: pull stuff like this which are maintenance refactors without intent changes into a separate PR for easier review and merge

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -40,9 +42,11 @@ const (
 	// attached to all token secrets this controller create
 	CreateDockercfgSecretsController = "openshift.io/create-dockercfg-secrets"
 
-	// PendingTokenAnnotation contains the name of the token secret that is waiting for the
+	DockercfgExpirationAnnotationKey = "openshift.io/dockercfg-token-expiry"
+
+	// PendingTokenAnnotation contains the name of the dockercfg secret that is waiting for the
 	// token data population
-	PendingTokenAnnotation = "openshift.io/create-dockercfg-secrets.pending-token"
+	PendingTokenAnnotation = "openshift.io/create-dockercfg-secrets.pending-secret"
```
deads2k, line 49: a change here looks suspicious.  I would expect a new constant instead.

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -394,199 +442,192 @@ func (e *DockercfgController) syncServiceAccount(key string) error {
...
+	// If this service account has no record of a pending token name, record one
+	if len(pendingTokenName) == 0 {
+		pendingTokenName = secret.Strategy.GenerateName(getDockercfgSecretNamePrefix(serviceAccount.Name))
+		if serviceAccount.Annotations == nil {
+			serviceAccount.Annotations = map[string]string{}
 		}
-		if !exists {
-			continue
+		serviceAccount.Annotations[PendingTokenAnnotation] = pendingTokenName
+		updatedServiceAccount, err := e.client.CoreV1().ServiceAccounts(serviceAccount.Namespace).Update(ctx, serviceAccount, metav1.UpdateOptions{})
```
deads2k, line 550: I recommend using the apply client for this.  You can do it as a separate step, but it will reduce conflicts.

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -394,199 +442,192 @@ func (e *DockercfgController) syncServiceAccount(key string) error {
...
+
+		blockDeletion := false
+		ownerRef := metav1.NewControllerRef(serviceAccount, v1.SchemeGroupVersion.WithKind("ServiceAccount"))
+		ownerRef.BlockOwnerDeletion = &blockDeletion
+		dockercfgSecret.SetOwnerReferences([]metav1.OwnerReference{*ownerRef})
+
+		klog.V(4).Infof("Creating dockercfg secret %q for service account %s/%s", dockercfgSecret.Name, serviceAccount.Namespace, serviceAccount.Name)
+
+		// Save the secret
+		_, err = e.client.CoreV1().Secrets(serviceAccount.Namespace).Create(ctx, dockercfgSecret, metav1.CreateOptions{})
```
deads2k, line 598: why create before getting the content?  With the tokenrequest API, the call to get the content is synchronous and it would simplify some of the waiting logic, right?

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -129,114 +139,93 @@ type DockercfgController struct {
...
+func runner(syncFunc func(ctx context.Context) bool) func(ctx context.Context) {
+	return func(ctx context.Context) {
+		for syncFunc(ctx) {
+		}
 	}
-	<-stopCh
 }
 
-func (c *DockercfgController) waitForDockerURLs(ready chan<- struct{}, stopCh <-chan struct{}) {
+func (c *DockercfgController) waitForDockerURLs(ctx context.Context, ready chan<- struct{}) {
```
deads2k, line 222: separate from this PR (before or after), I wonder if these are still logically configurable.  They used to be cyclical with the rest of the controller-manager so I built this logic, but I wonder if on openshift these are now practically fixed in a way that the names can be predicted during installation and reacted to if they ever change to avoid the interlock and self-discovery logic.

Just imagining future you trying to maintain this area of code.

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
```diff
@@ -394,199 +442,192 @@ func (e *DockercfgController) syncServiceAccount(key string) error {
...
-	for _, secretRef := range serviceAccount.Secrets {
-		secret, exists, err := e.secretCache.GetByKey(fmt.Sprintf("%s/%s", secretRef.Namespace, secretRef.Name))
-		if err != nil {
-			return err
+// createDockerPullSecret creates a dockercfg secret based on the token secret
+func (e *DockercfgController) createDockerPullSecret(ctx context.Context, serviceAccount *v1.ServiceAccount) (*v1.Secret, bool, error) {
+	pendingTokenName := serviceAccount.Annotations[PendingTokenAnnotation]
+
+	// If this service account has no record of a pending token name, record one
+	if len(pendingTokenName) == 0 {
```
deads2k, line 544: If you can build a valid dockerconfig secret synchronously, you'd be able to first build the secret and second directly update the serviceaccount with the pull secret and never need the intermediate annotation, right?

deads2k:
Just so I'm clear, this is about how we handle new dockercfg secrets, but how do we migrate existing ones and how do existing ones delete the old token secret and use a new tokenrequest based token instead?

pkg/serviceaccounts/controllers/create_dockercfg_secrets.go
soltysh: I was considering patch when looking through the code, this helped with with other controllers in the past.

stlaz:
/remove-lifecycle stale
I might be able to get to this soon-ish

PR #228, opened 2022-06-23
WIP: allow for future extraction of route controllers

mfojtik:
This pull request will:

* eliminate the k8s.io/kubernetes dependency from `pkg/route`
* remove `legacyschema` as the event recorder is not used/wired anyway

This should make it much easier to pull this package out (to library-go) and consume later in projects like microshift.

### DIFF
--- a/pkg/route/allocator/bitmap.go
+++ b/pkg/route/allocator/bitmap.go
@@ -1,0 +1,207 @@
+/*
+Copyright 2015 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


pkg/route/allocator/bitmap.go
```diff
@@ -0,0 +1,207 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
mfojtik, line 2: FYI, this was copied from upstream. It has not changed for months/years. It is a relatively small risk to keep copied vs. vendoring entire Kubernetes to get these. Eventually, we can try to extract these out upstream, but it will require changing the internal API to external API.

PR #230, opened 2022-07-20
Bug 2110617: Split route controller

ravisantoshgudimetla:


pkg/cmd/controller/route-controller/ingress.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
+	networkingv1 "k8s.io/client-go/kubernetes/typed/networking/v1"
+
+	routeclient "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	"github.com/openshift/openshift-controller-manager/pkg/route/ingress"
+)
+
+func RunIngressToRouteController(ctx *ControllerContext) (bool, error) {
```
deads2k, line 11: straight move?

pkg/cmd/controller/route-controller/interfaces.go
```diff
@@ -0,0 +1,234 @@
...
+	HighRateLimitClientBuilder ControllerClientBuilder
+
+	KubernetesInformers                informers.SharedInformerFactory
+	OpenshiftConfigKubernetesInformers informers.SharedInformerFactory
+
+	RouteInformers    routeinformer.SharedInformerFactory
+	ConfigInformers   configinformer.SharedInformerFactory
+	OperatorInformers operatorinformer.SharedInformerFactory
+
+	GenericResourceInformer genericinformers.GenericResourceInformer
```
deads2k, line 57: dead?

pkg/cmd/controller/route-controller/network.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/openshift/openshift-controller-manager/pkg/route/ingressip"
+)
+
+func RunIngressIPController(ctx *ControllerContext) (bool, error) {
```
deads2k, line 11: straight move?

pkg/cmd/openshift-controller-manager/cmd.go
```diff
@@ -131,5 +130,9 @@ func (o *OpenShiftControllerManager) RunControllerManager() error {
 	if err != nil {
 		return err
 	}
-	return RunOpenShiftControllerManager(config, clientConfig)
+	if err = RunOpenShiftControllerManager(config, clientConfig); err != nil {
+		return err
+	}
+	// TODO: Split RouteController into its own config and command.
+	return RunRouteControllerManager(config, clientConfig)
```
deads2k, line 137: instead of this, add `RunRouteControllerManager` to `ControllerInitializers` in `pkg/cmd/controller/config.go` and remove the lines for 

```go
	"openshift.io/ingress-ip":       RunIngressIPController,
	"openshift.io/ingress-to-route": RunIngressToRouteController,
```

This way, the main controller launcher will launch the logic that gets the lease and starts the route controllers.

pkg/cmd/controller/route-controller/route.go
```diff
@@ -0,0 +1,83 @@
...
+	eventBroadcaster.StartLogging(klog.Infof)
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	eventRecorder := eventBroadcaster.NewRecorder(legacyscheme.Scheme, v1.EventSource{Component: "cluster-route-controller"})
+	id, err := os.Hostname()
+	if err != nil {
+		return false, err
+	}
+	// Create a new lease for the route controller manager
+	rl, err := resourcelock.New(
+		"configmapsleases",
```
deads2k, line 56: just leases

pkg/cmd/controller/config.go
```diff
@@ -21,9 +21,9 @@ var ControllerInitializers = map[string]InitFunc{
 	"openshift.io/templateinstance":          RunTemplateInstanceController,
 	"openshift.io/templateinstancefinalizer": RunTemplateInstanceFinalizerController,
 
-	"openshift.io/unidling":         RunUnidlingController,
-	"openshift.io/ingress-ip":       RunIngressIPController,
```
deads2k, line 25: create a new map with just these

pkg/cmd/controller/route-controller/ingress.go
```diff
@@ -8,19 +8,19 @@ import (
 	"github.com/openshift/openshift-controller-manager/pkg/route/ingress"
 )
 
-func RunIngressToRouteController(ctx *ControllerContext) (bool, error) {
```
deads2k, line 11: this should be the same.  just take the routeControllerContext

pkg/cmd/controller/route-controller/network.go
```diff
@@ -8,22 +8,22 @@ import (
 	"github.com/openshift/openshift-controller-manager/pkg/route/ingressip"
 )
 
-func RunIngressIPController(ctx *ControllerContext) (bool, error) {
```
deads2k, line 11: this should be the same. just take the routeControllerContext

pkg/cmd/controller/route-controller/route.go
```diff
@@ -0,0 +1,83 @@
...
+			return false, err
+		}
+	}
+	if ctx.IsControllerEnabled("openshift.io/ingress-to-route") {
+		ingressToRouteController, err = getIngressToRouteController(ctx)
+		if err != nil {
+			return false, err
+		}
+	}
+	routeControllerManager := func(cntx context.Context) {
```
deads2k, line 38: this logic should be like originControllerManager below and initialize based on the new route controller init function map

pkg/cmd/controller/route-controller/route.go
```diff
@@ -0,0 +1,83 @@
...
+
+func RunRouteControllerManager(ctx *ControllerContext) (bool, error) {
+	kubeClient, err := ctx.ClientBuilder.Client(infraIngressToRouteControllerServiceAccountName)
+	if err != nil {
+		return true, err
+	}
+	config := ctx.OpenshiftControllerConfig
+	var ingressToRouteController *ingress.Controller
+	var ingressController *ingressip.IngressIPController
+	if ctx.IsControllerEnabled("openshift.io/ingress-ip") {
```
deads2k, line 26: I don't think this is necessary once you init based on the map

pkg/cmd/controller/route-controller/route.go
ravisantoshgudimetla: startControllers go ahead and start the individual controller, don't we want to start both the controllers in controller-manager code?

pkg/cmd/controller/route-controller/network.go
ravisantoshgudimetla: We're actually taking `OpenshiftControllerConfig.Ingress.IngressIPNetworkCIDR` here

pkg/cmd/controller/route.go
```diff
@@ -0,0 +1,97 @@
...
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().Events("")})
+	eventRecorder := eventBroadcaster.NewRecorder(legacyscheme.Scheme, v1.EventSource{Component: "cluster-route-controller"})
+	id, err := os.Hostname()
+	if err != nil {
+		return false, err
+	}
+	// Create a new lease for the route controller manager
+	rl, err := resourcelock.New(
+		"leases",
+		"openshift-route-controller-manager", // TODO: This namespace needs to be created by ocm for now.
```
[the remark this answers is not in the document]
deads2k: > by ocmo no?

I don't understand this comment.  Yes, the operator should create this namespace and could do so before this merges.

deads2k:
I think this PR does what I expect now.

pkg/cmd/controller/route.go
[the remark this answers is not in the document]
atiratree: I see Ravi opened a PR in the openshift-controller-manager-operator (ocmo), so everything should be in order

pkg/cmd/controller/route.go
```diff
@@ -0,0 +1,97 @@
...
+
+func RunRouteControllerManager(ctx *ControllerContext, clientConfig *rest.Config) (bool, error) {
+	kubeClient, err := ctx.ClientBuilder.Client(infraIngressToRouteControllerServiceAccountName)
+	if err != nil {
+		return true, err
+	}
+	config := ctx.OpenshiftControllerConfig
+
+	routeControllerManager := func(cntx context.Context) {
+		// Start Route Controllers
```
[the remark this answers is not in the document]
ravisantoshgudimetla: I think we can add it later. After looking at the code, I am a bit more confident that this is not needed.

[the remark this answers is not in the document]
deads2k: It's less important here because there is no discovery of what APIs exist.  I'm still fine adding it.

soltysh (approved):
/approve

PR #231, opened 2022-07-26
Bug 1962502: Create custom metrics to list unmanaged Routes and Ingresses without IngressClassName

suleymanakbas91:
This PR adds two custom metrics to count the number of ingresses without IngressClassName and the number of routes owned by ingresses no longer managed.
- `pkg/route/ingress/ingress.go`: Registers metrics to the KubeRegistry. Moves the logic to determine which ingresses are managed to a separate `ingressManaged` function. 
- `pkg/route/ingress/metrics.go`: Implements [Registrable](https://github.com/kubernetes/component-base/blob/master/metrics/registry.go#L95) interface so that the metrics can be registered to the KubeRegistry. Creates the metrics and implements the collector for the metrics. 
- `pkg/route/ingress/metrics_test.go`: Adds unit-test for the metrics.

### DIFF
--- a/pkg/route/ingress/ingress.go
+++ b/pkg/route/ingress/ingress.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/kubernetes/pkg/api/legacyscheme"
 
 	routev1 "github.com/openshift/api/route/v1"
@@ -96,6 +97,11 @@ type Controller struct {
 	// expectationDelay controls how long the controller waits to observe its
 	// own creates. Exposed only for testing.
 	expectationDelay time.Duration
+
+	// Prometheus metrics
+	metricsCreated    bool
+	metricsCreateOnce sync.Once
+	metricsCreateLock sync.RWMutex
 }
 
 // expectations track an upcoming change to a named resource related
@@ -251,6 +257,11 @@ func NewController(eventsClient kv1core.EventsGetter, routeClient routeclient.Ro
 		},
 	})
 
+	if !c.MetricsCreated() {
+		legacyregistry.MustRegister(c)
+		klog.Info("ingress-to-route metrics registered with prometheus")
+	}
+
 	return c
 }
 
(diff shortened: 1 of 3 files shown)


pkg/route/ingress/ingress.go
```diff
@@ -577,6 +570,36 @@ func hasIngressOwnerRef(owners []metav1.OwnerReference) (string, bool) {
 	return "", false
 }
 
+func (c *Controller) isIngressManaged(ingress *networkingv1.Ingress) (bool, error) {
```
brandisher, line 573: Nit: drop the `is` and call it `ingressManaged()` for consistency.

pkg/route/ingress/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+func (c *Controller) Create(v *semver.Version) bool {
+	c.metricsCreateOnce.Do(func() {
+		c.metricsCreateLock.Lock()
+		defer c.metricsCreateLock.Unlock()
+		c.metricsCreated = true
+	})
+	return c.MetricsCreated()
+}
```
brandisher, line 36: It doesn't look lik `v` is used in the function body. What's the use case for making it a method argument?

pkg/route/ingress/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (c *Controller) ClearState() {
+	c.metricsCreateLock.Lock()
+	defer c.metricsCreateLock.Unlock()
+	c.metricsCreated = false
+}
+
+func (c *Controller) FQName() string {
+	return routeController
+}
```
brandisher, line 50: I think this needs a comment with an explanation. I'm not sure what `FQ` stands for in this context.

pkg/route/ingress/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (c *Controller) Describe(ch chan<- *prometheus.Desc) {
+	unmanagedRoutes.Describe(ch)
+	ingressesWithoutClassName.Describe(ch)
+}
+
+func (c *Controller) Collect(ch chan<- prometheus.Metric) {
+	//ingress
```
brandisher, line 58: This comment doesn't add context; can it be removed?

pkg/route/ingress/metrics.go
```diff
@@ -0,0 +1,102 @@
...
+		labelVal := 0
+		if ingressInstance.Spec.IngressClassName == nil {
+			labelVal = 1
+			ingressesWithoutClassName.WithLabelValues(ingressInstance.Name).Set(float64(labelVal))
+		}
+	}
+
+	ingressesWithoutClassName.Collect(ch)
+
+	//route
```
brandisher, line 75: Same thing here; looks like this can be removed.

pkg/route/ingress/ingress.go
suleymanakbas91: Done.

pkg/route/ingress/metrics.go
suleymanakbas91: I improved the comments instead.

pkg/route/ingress/metrics.go
suleymanakbas91: I've added a comment.

quarterpin:
/label qe-approved

Verified via pre-merge workflow. More details can be found in : https://bugzilla.redhat.com/show_bug.cgi?id=1962502#c4

candita:
@suleymanakbas91 can you explain why was the decision made to add the metrics here instead of in cluster-ingress-operator?

atiratree:
@suleymanakbas91 fyi https://github.com/openshift/openshift-controller-manager/pull/240 removes these controllers from OCM repo. Any future PRs should go to https://github.com/openshift/route-controller-manager from now on

suleymanakbas91:
Moving this to route-controller-manager: https://github.com/openshift/route-controller-manager/pull/8

candita:
I will reassign myself and Ryan to the new PR

PR #234, opened 2022-08-05
Bug 2117235: separate route controllers to a new command

atiratree:
TODO:
- [x] https://github.com/openshift/openshift-controller-manager/pull/235

### DIFF
--- a/cmd/route-controller-manager/main.go
+++ b/cmd/route-controller-manager/main.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"os"
+	"runtime"
+
+	"github.com/openshift/library-go/pkg/serviceability"
+	"github.com/spf13/cobra"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/component-base/cli"
+
+	route_controller_manager "github.com/openshift/openshift-controller-manager/pkg/cmd/route-controller-manager"
+	"github.com/openshift/openshift-controller-manager/pkg/routeversion"
+)
+
+func main() {
+	stopCh := genericapiserver.SetupSignalHandler()
+
+	defer serviceability.BehaviorOnPanic(os.Getenv("OPENSHIFT_ON_PANIC"), routeversion.Get())()
+	defer serviceability.Profile(os.Getenv("OPENSHIFT_PROFILE")).Stop()
+
+	if len(os.Getenv("GOMAXPROCS")) == 0 {
+		runtime.GOMAXPROCS(runtime.NumCPU())
+	}
+
+	command := NewRouteControllerManagerCommand(stopCh)
+	code := cli.Run(command)
+	os.Exit(code)
+}
+
+func NewRouteControllerManagerCommand(stopCh <-chan struct{}) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "route-controller-manager",
+		Short: "Command for additional management of ingress and Route resources",
+		Run: func(cmd *cobra.Command, args []string) {
+			cmd.Help()
+			os.Exit(1)
+		},
+	}
+	start := route_controller_manager.NewRouteControllerManagerCommand("start", os.Stdout, os.Stderr, stopCh)
+	cmd.AddCommand(start)
+
+	return cmd
+}
--- a/pkg/cmd/controller/route.go
+++ b/pkg/cmd/controller/route.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/leaderelection"
@@ -13,20 +14,16 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/api/legacyscheme"
 
+	openshiftcontrolplanev1 "github.com/openshift/api/openshiftcontrolplane/v1"
+
 	routecontrollers "github.com/openshift/openshift-controller-manager/pkg/cmd/controller/route"
 )
 
-func RunRouteControllerManager(ctx *ControllerContext, clientConfig *rest.Config) (bool, error) {
-	kubeClient, err := ctx.ClientBuilder.Client(infraIngressToRouteControllerServiceAccountName)
-	if err != nil {
-		return true, err
-	}
-	config := ctx.OpenshiftControllerConfig
-
+func RunRouteControllerManager(config *openshiftcontrolplanev1.OpenShiftControllerManagerConfig, kubeClient kubernetes.Interface, clientConfig *rest.Config) (bool, error) {
 	routeControllerManager := func(cntx context.Context) {
 		// Start Route Controllers
 		// TODO: This can be split further
-		routeControllerContext, err := routecontrollers.NewControllerContext(cntx, config, clientConfig)
+		routeControllerContext, err := routecontrollers.NewControllerContext(cntx, *config, clientConfig)
 		if err != nil {
 			klog.Fatal(err)
 		}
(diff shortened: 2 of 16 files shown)


pkg/cmd/controller/route/apiserver_authenticator.go
```diff
@@ -0,0 +1,52 @@
+package route
```
atiratree, line 1: to achieve removal of openshift-controller-manager imports I had to duplicate a lot of files (mainly for the server and genericinformers)

- Is this necessary? I think we could split repo even with files that are used by both controllers.
- since it is a duplicate copy the history is preserved, but there seems to be ways on how to achive this (https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history). Please let me know if we should go this way
- also most of these files will be deleted anyway after the final refactoring

cmd/route-controller-manager/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+	"runtime"
+
+	"github.com/openshift/library-go/pkg/serviceability"
+	"github.com/spf13/cobra"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/component-base/cli"
+
+	route_controller_manager "github.com/openshift/openshift-controller-manager/pkg/cmd/route-controller-manager"
+	"github.com/openshift/openshift-controller-manager/pkg/version"
```
atiratree, line 13: this is the last file that is not duplicated, but I can duplicate that as well.

pkg/cmd/controller/route.go
```diff
@@ -13,20 +14,16 @@ import (
...
 )
 
-func RunRouteControllerManager(ctx *ControllerContext, clientConfig *rest.Config) (bool, error) {
-	kubeClient, err := ctx.ClientBuilder.Client(infraIngressToRouteControllerServiceAccountName)
-	if err != nil {
-		return true, err
-	}
-	config := ctx.OpenshiftControllerConfig
-
+func RunRouteControllerManager(config *openshiftcontrolplanev1.OpenShiftControllerManagerConfig, kubeClient kubernetes.Interface, clientConfig *rest.Config) (bool, error) {
```
atiratree, line 22: this file is in a shared package. Should I move this to a route package so everything is also split on a package level?

pkg/client/routegenericinformers/interface.go
```diff
@@ -0,0 +1,80 @@
...
+
+type GenericResourceInformer interface {
+	ForResource(resource schema.GroupVersionResource) (informers.GenericInformer, error)
+	Start(stopCh <-chan struct{})
+}
+
+// GenericInternalResourceInformerFunc will return an internal informer for any resource matching
+// its group resource, instead of the external version. Only valid for use where the type is accessed
+// via generic interfaces, such as the garbage collector with ObjectMeta.
+type GenericInternalResourceInformerFunc func(resource schema.GroupVersionResource) (informers.GenericInformer, error)
```
deads2k, line 19: which part of the route controller uses this?

pkg/cmd/controller/route.go
deads2k: Yes, split this package.  I suggest creating `pkg/cmd/routecontroller`

cmd/route-controller-manager/main.go
deads2k: duplicate.

pkg/cmd/route-controller-manager/controller_manager.go
```diff
@@ -0,0 +1,36 @@
...
+
+	openshiftcontrolplanev1 "github.com/openshift/api/openshiftcontrolplane/v1"
+	"github.com/openshift/library-go/pkg/serviceability"
+
+	origincontrollers "github.com/openshift/openshift-controller-manager/pkg/cmd/controller"
+	routecontrollers "github.com/openshift/openshift-controller-manager/pkg/cmd/controller/route"
+	"github.com/openshift/openshift-controller-manager/pkg/version"
+)
+
+func RunrouteControllerManager(config *openshiftcontrolplanev1.OpenShiftControllerManagerConfig, clientConfig *rest.Config) error {
```
deads2k, line 16: RunRouteControllerManager

pkg/client/routegenericinformers/interface.go
atiratree: nice catch. As far as I can tell, it is a dead code and is not used. This applies to the openshift-controller-manager as well. 

- I can remove this whole file
- I can do a followup and remove it from the main controller as well

pkg/cmd/controller/route/apiserver_authenticator.go
atiratree: https://github.com/openshift/openshift-controller-manager/pull/234#discussion_r940186953 has shown we do not need the genericinformers so this concerns only the server code.

since we will be deleting the server code later, I am for not preserving the history of the server files and doing just plain copy.

pkg/client/routegenericinformers/interface.go
deads2k: The generic informer itself is not used by the route controllers.  The route controllers use specific informers.

ingvagabund: > This applies to the openshift-controller-manager as well.

> I can do a followup and remove it from the main controller as well

I was referring to the above comments about the main controller. I will wait for the changes to see the final outcome :)

pkg/client/routegenericinformers/interface.go
```diff
@@ -0,0 +1,80 @@
+package genericinformers
```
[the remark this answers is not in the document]
atiratree: removed

pkg/client/routegenericinformers/interface.go
atiratree: all of the code related to generic informers will be removed, even the `ControllerContext.ToGenericInformer()` invocation

atiratree:
/retest
runs ok locally, seems unrelated

pkg/routeversion/version.go
```diff
@@ -0,0 +1,74 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
ingvagabund, line 2: Nit: s/2017/2022

pkg/routeversion/version.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"runtime"
+
+	"k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+
+	"k8s.io/apimachinery/pkg/version"
+)
+
+var (
```
ingvagabund, line 29: None of these variables get populated given the [build-machine-go assumes](https://github.com/openshift/openshift-controller-manager/blob/master/vendor/github.com/openshift/build-machinery-go/make/lib/golang.mk#L69) only `$(GO_PACKAGE)/pkg/version` path. Nevertheless, once the code is moved into the new repository, the version path gets normalized and properly populated again.

pkg/cmd/route-controller-manager/controller_manager.go
```diff
@@ -0,0 +1,36 @@
...
+func RunRouteControllerManager(config *openshiftcontrolplanev1.OpenShiftControllerManagerConfig, clientConfig *rest.Config) error {
+	serviceability.InitLogrusFromKlog()
+	kubeClient, err := kubernetes.NewForConfig(clientConfig)
+	if err != nil {
+		return err
+	}
+
+	// only serve if we have serving information.
+	if config.ServingInfo != nil {
+		klog.Infof("Starting controllers on %s (%s)", config.ServingInfo.BindAddress, routeversion.Get().String())
```
ingvagabund, line 25: Even though `routeversion` is not properly populated and the version string contains `unknown` keyword, `RunOpenShiftControllerManager` still prints the right version.

pkg/cmd/route-controller-manager/cmd.go
```diff
@@ -0,0 +1,136 @@
...
+	configFileLocation := path.Dir(absoluteConfigFile)
+
+	config := obj.(*openshiftcontrolplanev1.OpenShiftControllerManagerConfig)
+	/// this isn't allowed to be nil when by itself.
+	// TODO remove this when the old path is gone.
+	if config.ServingInfo == nil {
+		config.ServingInfo = &configv1.HTTPServingInfo{}
+	}
+	// TODO: resolve split from OCM
+	config.ServingInfo.BindAddress = "0.0.0.0:9443"
```
ingvagabund, line 125: Where does `9443` port number come from?

pkg/routeversion/version.go
atiratree: it is a copy of already existing class, but I guess we can put 2022 since this is a common generic code

pkg/cmd/route-controller-manager/cmd.go
atiratree: I have been using this for testing, eg. running this in the same pod as openshift-controller-manager, but since we will be moving this to a separate pod and ns I can put back the default 8443

pkg/routeversion/version.go
atiratree: ack, I think we can live with the broken version for the time of the move

soltysh (approved):
/approve

PR #238, opened 2022-09-21
remove route controllers from OCM process (openshift-controller-manager start)

atiratree:
and run them only separately via [openshift-controller-manager route-controller-manager-start](https://github.com/openshift/openshift-controller-manager/blob/master/cmd/openshift-controller-manager/main.go#L69)


soltysh (approved):
/lgtm
/approve

PR #245, opened 2022-10-11
WIP: HostUsers=false for unprivileged build pods + UserNamespaces

nalind:
If the `UserNamespacesStatelessPodsSupport` feature is enabled (known in 1.25 and later, since https://github.com/kubernetes/kubernetes/pull/111090), set `HostUsers=false` for unprivileged build pod specs, in addition to continuing to set the non-standardized annotation that older versions of CRI-O could be configured to accept instead.


nalind:
Yep, holler if it needs a rebase.

coreydaley:
/assign @coreydaley 
for lgtm

coreydaley:
@nalind It looks like you need to setup a bug in JIRA for this to be merged (and possibly backported?)

nalind:
> @nalind It looks like you need to setup a bug in JIRA for this to be merged (and possibly backported?)

Created BUILD-596.  It won't make sense on anything earlier than 4.12, since 4.11 was based on Kubernetes 1.24, which doesn't know about the feature.

nalind:
@coreydaley PTAL

pkg/cmd/controller/build.go
```diff
@@ -55,12 +56,16 @@ func RunBuildController(ctx *ControllerContext) (bool, error) {
...
 	fg := ctx.OpenshiftControllerConfig.FeatureGates
 	csiVolumesEnabled := false
+	hostUsersForUserNS := false
 	if fg != nil {
 		for _, v := range fg {
 			v = strings.TrimSpace(v)
 			if v == "BuildCSIVolumes=true" {
 				csiVolumesEnabled = true
 			}
+			if v == string(features.UserNamespacesStatelessPodsSupport)+"=true" {
```
coreydaley, line 66:
```suggestion
			if v == fmt.Sprintf("%s=true", features.UserNamespacesStatelessPodsSupport) {
```

coreydaley (changes requested):

pkg/cmd/controller/build.go
nalind: Making this change.

nalind:
@coreydaley PTAL

nalind:
Holding because https://kubernetes.io/blog/2023/09/13/userns-alpha/ notes that the feature gate is being renamed in 1.28, and I don't think this is getting merged until 4.15 at least.
/hold

PR #246, opened 2022-11-14
update the deploy pod to provide failure in pod

deads2k:
While debugging a deployer pod problem, I couldn't find the message I needed.  This leaves the current file location alone and it will continue to be used if present, but without it, the sysout will be used.

/assign @soltysh


deads2k:
I need this to figure out why https://github.com/openshift/oc/pull/1288 fails to figure out why https://github.com/openshift/kubernetes/pull/1393 fails

soltysh (approved):
/lgtm
/approve
/hold
for the verify fix, you need `make update-gofmt`

deads2k:
fixed gofmt, relabelling

/hold cancel

PR #252, opened 2023-02-18
BUILD-558: Easier to Use Buildpacks on OpenShift Builds (Custom Strategy)

otaviof:
Custom Build Strategy (for Buildpacks)
--------------------------------------

This pull request changes the `BuildConfig` Custom strategy ([BUILD-558](https://issues.redhat.com//browse/BUILD-558)) to make simpler to run a custom party image builder.

The [testing](#testing) section shows how to employ a [Cloud Native Builder (CNB)][customBuilder], which still needs to [intervene and prepare the container][customBuilderEntrypoint] before being able to run the build lifecycle process.

The section ["Next Steps"](#next-steps) shares some ideas for the upcoming interactions to improve the developer experience.

# Pull-Request Changes

## Git-Clone Init-Container

Following `Source` and `Docker` strategies, the source-code is cloned using a init-container and exposed in the build POD using a `emptyDir: {}` volume, shared between containers.

### Writable Working Directory

Additionally, the `workingDir` is set to `/tmp/build/inputs` and the directory is made writable, the location used for the Git-Clone process is a shared `emptyDir: {} ` mount.

The [example application][otaviofNodejsEx] is based on Node.js meaning during the regular build process the directories `node_modules` will be created. It's a common use case having the (temporary) working directory on read/write mode.

## `OUTPUT_REGISTRY_IMAGE`

The fully qualified output image name is informed by the Build controller to the custom builder, a small modification to avoid it on the custom builder process.

# Next Steps

Sharing some ideas out loud to gather early feedback.

## Script Section

At least a few tweaks will have to take place before being able to invoke the custom builder, like for instance, when [using the internal container registry with self signed certificates][customBuilderCATrust] the builder images needs to add the CA to the local trust.

The most simplistic approach to Paketo's CNB, taking in considering this PR, is:

```yaml
---
apiVersion: build.openshift.io/v1
kind: BuildConfig
spec:
  strategy:
    type: Custom
    customStrategy:
      script: |
        /cnb/lifecycle/creator -app="." "${OUTPUT_REGISTRY_IMAGE}"
```

The `.spec.strategy.customStrategy.script` helps the users to setup the build enviornment and empower using third party tools directly on the `BuildConfig` resource.

## Deprecate `exposeDockerSocket`

We should stop allowing the access to the node's docker-socket by all means, the attribute `.spec.strategy.customStrategy.exposeDockerSocket` should be deprecated as soon as possible.

OpenShift should focus the Custom strategy to non-privileged builds only.

## Unprivileged Builds by Default

OpenShift Builds Custom strategy should have non-privileged by default, the `BUILD_PRIVILEGED` environment variable should have its default set to `false`.

This improves the developer experience and also makes the recommended approach easier to addopt.

## Custom Strategy on `new-app` Workflow

Currently `oc new-app` (or `new-build`) do not support Custom strategy, we should improve the command line tool to support this style of builds as well. For instance:

```bash
oc new-app docker.io/paketobuildpacks/builder:base~https://github.com/otaviof/nodejs-ex.git \
    --name="nodejs-ex" \
    --strategy="custom" \
    --env="CNB_PLATFORM_API=0.10" \
    --custom-script='/cnb/lifecycle/creator -app="." "${OUTPUT_REGISTRY_IMAGE}"'
```

The most important aspects in this example are related to the ability to define the `.spec.strategy.customStrategy.from.name` and `.spec.strategy.customStrategy.script` attributes.

Most of the heavy lifting is already in place.

# Testing

## Buildpacks CNB

The [`os-custom-paketo-builder` image][customBuilder] is the example implementation of a custom Buildpacks CNB for OpenShift, it takes care of setting the stage before running the Buildpacks Lifecycle.

Make sure the `ImageStream` is available by following the [repository documentation][customBuilder].

## `BuildConfig`

Follow the steps in the [custom CNB image repository][customBuilderOpenShift] to create the `ImageStreams`, configure the registry authentication and allow `anyuid` SCC.

Then, apply the `BuildConfig` resource:

```bash
oc apply -f https://github.com/otaviof/os-custom-paketo-builder/blob/main/os/buildconfig.yaml
```

Finally, start the build:

```bash
oc start-build nodejs-ex --follow --wait
```

The Node.js application build is going to be ready for [deploying it too][customBuilderDeployment].

[customBuilder]: https://github.com/otaviof/os-custom-paketo-builder/
[customBuilderCATrust]: https://github.com/otaviof/os-custom-paketo-builder/blob/01a9291577b358fdc6fb70a952936073e6649883/build.sh#L49-L60
[customBuilderDeployment]: https://github.com/otaviof/os-custom-paketo-builder/blob/main/os/README.md#deploying
[customBuilderEntrypoint]: https://github.com/otaviof/os-custom-paketo-builder/blob/main/build.sh
[customBuilderOpenShift]: https://github.com/otaviof/os-custom-paketo-builder/blob/main/os/README.md
[otaviofNodejsEx]: https://github.com/otaviof/nodejs-ex

### DIFF
--- a/pkg/build/controller/strategy/custom.go
+++ b/pkg/build/controller/strategy/custom.go
@@ -3,6 +3,7 @@ package strategy
 import (
 	"errors"
 	"fmt"
+	"path"
 
 	"k8s.io/klog/v2"
 
@@ -30,6 +31,7 @@ func init() {
 
 // CustomBuildStrategy creates a build using a custom builder image.
 type CustomBuildStrategy struct {
+	Image string // git clone init-container image
 }
 
 // CreateBuildPod creates the pod to be used for the Custom build
@@ -63,6 +65,10 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 	}
 	addTrustedCAMountEnvVar(build.Spec.MountTrustedCA, &containerEnv)
 
+	if len(strategy.Env) > 0 {
+		buildutil.MergeTrustedEnvWithoutDuplicates(strategy.Env, &containerEnv, true)
+	}
+
 	if build.Spec.Output.To != nil {
 		addOutputEnvVars(build.Spec.Output.To, &containerEnv)
 		if err != nil {
@@ -89,6 +95,7 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 	}
 
 	securityContext := securityContextForBuild(strategy.Env)
+	workingDir := path.Join(buildutil.BuildWorkDirMount, "inputs")
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      buildutil.GetBuildPodName(build),
@@ -104,11 +111,42 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 					Env:                      containerEnv,
 					SecurityContext:          securityContext,
 					TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+					VolumeMounts: []corev1.VolumeMount{{
+						Name:      BuildWorkDirVolume,
+						MountPath: buildutil.BuildWorkDirMount,
+					}, {
+						Name:      "buildcachedir",
+						MountPath: buildutil.BuildBlobsMetaCache,
+					}},
+					Resources: build.Spec.Resources,
+					// setting the container to run directly in the location the repository is cloned
+					// using the openshift helper
+					WorkingDir: workingDir,
 				},
 			},
+			Volumes: []corev1.Volume{{
+				Name: BuildWorkDirVolume,
+				VolumeSource: corev1.VolumeSource{
+					EmptyDir: &corev1.EmptyDirVolumeSource{},
+				},
+			}, {
+				Name: "buildcachedir",
+				VolumeSource: corev1.VolumeSource{
+					HostPath: &corev1.HostPathVolumeSource{
+						Path: buildutil.BuildBlobsMetaCache,
+					},
+				},
+			}},
 			RestartPolicy: corev1.RestartPolicyNever,
 			NodeSelector:  build.Spec.NodeSelector,
 		},
+	}
+
+	if build.Spec.Source.Git != nil || build.Spec.Source.Binary != nil {
+		// when source repository is declared, executing the clone as usual
+		setupGitCloneInitContainer(pod, build, bs.Image, containerEnv, securityContext)
+		// making sure the build working directory is writtable
+		setupChmodInitContainer(pod, build, bs.Image, containerEnv, securityContext, 0o777, workingDir)
 	}
 
 	pod = setupActiveDeadline(pod, build)
@@ -118,11 +156,6 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 	} else {
 		klog.V(2).Infof("ForcePull is enabled for %s build", build.Name)
 		pod.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
-	}
-	pod.Spec.Containers[0].Resources = build.Spec.Resources
-	if build.Spec.Source.Binary != nil {
-		pod.Spec.Containers[0].Stdin = true
-		pod.Spec.Containers[0].StdinOnce = true
 	}
 
 	if strategy.ExposeDockerSocket {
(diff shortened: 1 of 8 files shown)


otaviof:
@coreydaley, @jkhelil please consider.

otaviof:
/retitle BUILD-558: Easier to Adopt Custom Strategy on Buildpacks CNB

otaviof:
/retitle BUILD-558: Easier to Use Buildpacks on OpenShift Builds (Custom Strategy)

otaviof:
/retest

```
Feb 20 09:48:36.949: Some cluster operators are not ready: cert-manager
```

coreydaley:
/hold
We don't want this to merge

pkg/build/controller/strategy/util.go
```diff
@@ -765,6 +772,66 @@ func setupBuildVolumes(pod *corev1.Pod, buildVolumes []buildv1.BuildVolume) erro
...
+	containerEnv []corev1.EnvVar,
+	securityContext *corev1.SecurityContext,
+	mode int,
+	directory string,
+) {
+	chownContainer := corev1.Container{
+		Name:                     "git-chown",
+		Image:                    gitCloneImage,
+		Command:                  []string{"chmod"},
+		Args:                     []string{fmt.Sprintf("0%d", mode), directory},
```
nalind, line 790: Should this be
```suggestion
		Args:                     []string{fmt.Sprintf("0%o", mode), directory},
```

pkg/build/controller/strategy/custom.go
```diff
@@ -104,13 +107,33 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
...
 			NodeSelector:  build.Spec.NodeSelector,
 		},
 	}
 
+	// when source repository is declared, executing the clone as usual
+	if build.Spec.Source.Git != nil || build.Spec.Source.Binary != nil {
+		setupGitCloneInitContainer(pod, build, bs.Image, containerEnv, securityContext)
+	}
+	// making sure the build working directory is writtable during the process
+	setupChmodInitContainer(pod, build, bs.Image, containerEnv, securityContext, 777, workingDir)
```
nalind, line 135:
```suggestion
	setupChmodInitContainer(pod, build, bs.Image, containerEnv, securityContext, 0o777, workingDir)
```
to go with the recent change at util.go:790.

pkg/build/controller/strategy/custom.go
```diff
@@ -104,13 +107,33 @@ func (bs *CustomBuildStrategy) CreateBuildPod(build *buildv1.Build, additionalCA
 					Env:                      containerEnv,
 					SecurityContext:          securityContext,
 					TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+					VolumeMounts: []corev1.VolumeMount{{
+						Name:      BuildWorkDirVolume,
+						MountPath: buildutil.BuildWorkDirMount,
+					}},
+					// setting the container to run directly in the location the repository is cloned
+					// using the openshift helper
+					WorkingDir: workingDir,
```
nalind, line 116: Note to self: cri-o creates the container's working directory if necessary.

pkg/build/controller/strategy/custom_test.go
```diff
@@ -57,20 +57,9 @@ func TestCustomCreateBuildPod(t *testing.T) {
...
-	// expected volumes:
-	// docker socket
-	// push secret
-	// source secret
-	// additional secrets
-	// build-system-configmap
-	// certificate authorities
-	// container storage
-	// container run
-	// global CA injection configmap
```
nalind, line 69: Why delete this comment instead of updating it?  The tests for the docker and source strategies will still include similar comments.

pkg/build/controller/strategy/custom.go
otaviof: That's good to know, thanks for sharing!

pkg/build/controller/strategy/custom_test.go
otaviof: Oops, I'll revert and update the comments accordingly.

pkg/build/controller/strategy/custom.go
nalind: Oh, the chmod container probably needs this set, too, so that the runtime will ensure the directory exists.  If the git-clone container doesn't create the directory, or it isn't used, the runtime won't know to create it for the chmod container otherwise.

otaviof: Indeed, the `git-chmod` container definitely needs that setting, thanks for the heads up!

PR #254, opened 2023-03-21
OCPBUGS-10588: mount build.Spec.Source.ConfigMaps for custom builder images

nalind:
When using a custom build strategy, mount config maps which were listed as sources, like we do for docker and sti builds.


coreydaley:
@nalind Is this something that we should backport? Seems like a simple change that might be useful to have in previous versions as long as there are no conflicts.

nalind:
Does it help the original reporter if we do so?  The card was opened for 4.10, so we'd have to take it back through multiple releases.

coreydaley:
Yes.
If we can we should take it back as far as we can (but stop at/including 4.10)

PR #276, opened 2023-10-27
API-1644: Use bound service account tokens when generating pull secrets.

sanchezl:
Refactor the **openshift.io/serviceaccount-pull-secrets** controllers to use bound tokens when creating the image pull secrets for use with the internal image registry.

#### Controller Sync:
1. Add `openshift.io/internal-registry-pull-secret-ref` annotation to **ServiceAccount**.[^1]
2. Ensure **Secret** referenced in the annotation exists.[^2][^2a]
3. If contents of **Secret** are not valid (e.g. corrupt, expired, etc...), re-create.[^3][^3a]
4. Ensure **ServiceAccount.ImagePullSecrets** references the **Secret**.

#### To Do
- [x] https://github.com/openshift/openshift-apiserver/pull/403.
- [x] https://github.com/openshift/openshift-apiserver/pull/408.
- [x] Handle upgrade scenario.

[^1]: This will make the image pull secret name 'static' when compared to the previous implementation of this controller.
[^2]: An "empty" image pull secret is created at this step as it will need to already exist in order to serve as the object the token will be bound to.
[^2a]: No annotations needed on the Secret.
[^3]: An image pull secret is valid if it contains the correct number of entries/urls and the tokens are not expired or about to expire.
[^3a]: The actual token is examined for expiry determination.


pkg/internalregistry/legacy.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	// remove image pull secret legacy annotations
+	imagePullSecret, err = c.cleanupLegacyImagePullSecretAnnotations(ctx, imagePullSecret)
+	if err != nil {
+		return err
+	}
+
+	// remove the image pull secret from the service account's mountable secrets
+	c.cleanupLegacySecretRefs(ctx, imagePullSecret.Name, serviceAccount)
```
sanchezl, line 95: I misunderstood what the [secrets refs were for](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/service-account-v1/#:~:text=secrets%20([]ObjectReference)). I do not think I should be removing it.

pkg/internalregistry/controller.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	// get generated image pull secret for the service account
+	imagePullSecret, err := c.imagePullSecretForServiceAccount(ctx, serviceAccount)
+	if err != nil {
+		return err
+	}
+
+	// update registry authentication in the generated image pull secret
+	err = c.syncImagePullSecret(ctx, imagePullSecret, serviceAccount)
```
deads2k, line 111: service accounts can have more than one imagepullsecret, this appears to be the only callsite. Perhaps a controller that inspects image pull secrets would be robust

pkg/internalregistry/controller.go
```diff
@@ -0,0 +1,333 @@
...
+func newImagePullSecretForServiceAccount(serviceAccount *corev1.ServiceAccount) *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            serviceAccount.Annotations[InternalRegistryImagePullSecretRefKey],
+			Namespace:       serviceAccount.Namespace,
+			OwnerReferences: []metav1.OwnerReference{asOwnerReference(serviceAccount)},
+		},
+		Type: corev1.SecretTypeDockercfg,
+		Data: map[string][]byte{
+			corev1.DockerConfigKey: []byte("{}"),
```
deads2k, line 158: Why not initialize with the bound SA token already embedded so that we never go through a "created but not valid" state?

sanchezl: I cannot bind the token to an object (the image pull secret) that does not exist. Alternatively, I could not bind it at all, making it only time-bound.

sanchezl:
https://github.com/openshift/openshift-controller-manager/pull/288

PR #292, opened 2024-03-21
OCPBUGS-31213: Unable to look up the service account secrets for build

sanchezl:
Build is looking for image pull secrets in the **ServiceAccount.Secrets** field, when it should be looking in the **ServiceAccount.ImagePullSecrets** field instead. 

In 4.16 we will stop putting image pull secrets in the **ServiceAccount.Secrets** field.

### DIFF
--- a/pkg/build/buildutil/util.go
+++ b/pkg/build/buildutil/util.go
@@ -201,7 +201,7 @@ func FetchServiceAccountSecrets(secretStore v1lister.SecretLister, serviceAccoun
 	if err != nil {
 		return result, fmt.Errorf("Error getting push/pull secrets for service account %s/%s: %v", namespace, serviceAccount, err)
 	}
-	for _, ref := range sa.Secrets {
+	for _, ref := range sa.ImagePullSecrets {
 		secret, err := secretStore.Secrets(namespace).Get(ref.Name)
 		if err != nil {
 			continue
--- a/pkg/build/controller/build/build_controller_test.go
+++ b/pkg/build/controller/build/build_controller_test.go
@@ -2062,7 +2062,7 @@ func fakeKubeExternalClientSet(objects ...runtime.Object) kubernetes.Interface {
 	builderSA := &corev1.ServiceAccount{}
 	builderSA.Name = "builder"
 	builderSA.Namespace = "namespace"
-	builderSA.Secrets = []corev1.ObjectReference{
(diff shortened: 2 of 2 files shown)


adambkaplan (approved):
/approve

Before adding LGTM, I want to do some due diligence on docs/user impact. 

@sanchezl can you please point me to where this change is documented upstream? Is there any impact to how secret linking works for service accounts? See [OCP BuildConfig docs](https://docs.openshift.com/container-platform/4.15/cicd/builds/creating-build-inputs.html#builds-docker-credentials-private-registries_creating-build-inputs).

I also wonder if layered projects are aware this change is coming, such as Tekton, Shipwright, Argo, etc.

sanchezl:
> can you please point me to where this change is documented upstream? Is there any impact to how secret linking works for service accounts? See [OCP BuildConfig docs](https://docs.openshift.com/container-platform/4.15/cicd/builds/creating-build-inputs.html#builds-docker-credentials-private-registries_creating-build-inputs).

This PR checks for the presence of an image pull secret in the **.ImagePullSecrets** list, which seems appropriate even if we decide to retain image pull secrets in the **.Secrets** field[^1]. @adambkaplan, I believe we can continue discussing the removal of the image pull secret from the **.Secrets** field without delaying approval of this PR.

OpenShift documentation on linking image pull secrets to service accounts directs users to [specify the link type as **--for=pull**][2]. Without this pull request, the **FetchServiceAccountSecrets** function would not be able to find the image pull secret, as that option only adds the image pull secret to the list of [**.ImagePullSecrets**, not to the **.Secrets**][3].

[2]: https://docs.openshift.com/container-platform/4.15/openshift_images/managing_images/using-image-pull-secrets.html#:~:text=%24-,oc%20secrets%20link,-default%20%3Cpull_secret_name%3E
[3]: https://github.com/openshift/oc/blob/508e3d0e0aa2fe06e7b31552e29f43fb528fa599/pkg/cli/secrets/link.go#L131-L146

[^1]: My understanding is that the `.Secrets` field is only needed when a ServiceAccount has the **kubernetes.io/enforce-mountable-secrets** annotation set to **true**, in which case pods using the ServiceAccount can only access the secrets listed in that field. Additionally, upstream service account controllers use it to manage the legacy service account token secrets, which are expected to all be listed in the field.

PR #302, opened 2024-05-03
OCPBUGS-33241: panic in image trigger controller

sanchezl:
Handle getting a DeletedFinalStateUnknown instead of a runtime Object


soltysh (approved):
/approve

PR #304, opened 2024-05-12
Revert #288 "API-1644: Use bound service account tokens when generating pull secrets"

neisw:
Reverts #288 ; tracked by https://issues.redhat.com/browse/TRT-1665

Per [OpenShift policy](https://github.com/openshift/enhancements/blob/master/enhancements/release/improving-ci-signal.md#quick-revert), we are reverting this breaking change to get CI and/or nightly payloads flowing again.

Multiple payload failures due to hypershift conformance test failures

To unrevert this, revert this PR, and layer an additional separate commit on top that addresses the problem. Before merging the unrevert, please run these jobs on the PR and check the result of these jobs to confirm the fix has corrected the problem:

```
/payload-job periodic-ci-openshift-hypershift-release-4.16-periodics-e2e-aws-ovn-conformance
```

CC: @sanchezl

<div align="right">
PR created by <a href="https://github.com/stbenjam/revertomatic">Revertomatic<sup>:tm:</sup></a>
</div>

PR #305, opened 2024-05-13
OCPBUGS-33600: revert revert and fix for hypershift

sanchezl:
- Revert https://github.com/openshift/openshift-controller-manager/pull/304
- Add fallback to alternate method of obtaining token signing key hash

On Hypershift, the image pull secrets were created, but not properly initialized due to the `image_pull_secret_controller` getting stuck waiting for the existence of the **bound-service-account-signing-key** secret in the **openshift-kube-apiserver** namespace.

Ideally, Hypershift would prefer to mount the service account signing key as as a volume on OCM pod. This would match what we already do this for the KCM pod. This PR provides an OCM-only fix to get us going until we make the changes to the way hypershift runs OCM.

Since what we need is the just the hash of the service account signing public key, if OCM does not find the **bound-service-account-signing-key** secret on startup, OCM will create a throwaway API token and extract the service account signing public key's hash from the token.

Currently, on hypershift, the service account signing public key is provided as a CLI option. OCM would need to be restarted if the service account signing public key changes,

### DIFF
--- a/pkg/internalregistry/controllers/keyid_observation_controller.go
+++ b/pkg/internalregistry/controllers/keyid_observation_controller.go
@@ -9,11 +9,15 @@ import (
 	"time"
 
 	"golang.org/x/exp/slices"
+	"gopkg.in/square/go-jose.v2/jwt"
+	authenticationv1 "k8s.io/api/authentication/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	informers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
 	listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/keyutil"
@@ -22,14 +26,16 @@ import (
 )
 
 type keyIDObservation struct {
+	client     kubernetes.Interface
 	secrets    listers.SecretLister
 	cacheSyncs []cache.InformerSynced
 	queue      workqueue.RateLimitingInterface
 	ch         kidsChan
 }
 
-func NewKeyIDObservationController(secrets informers.SecretInformer, ch kidsChan) *keyIDObservation {
+func NewKeyIDObservationController(kubeclient kubernetes.Interface, secrets informers.SecretInformer, ch kidsChan) *keyIDObservation {
 	c := &keyIDObservation{
+		client:     kubeclient,
 		secrets:    secrets.Lister(),
 		ch:         ch,
 		cacheSyncs: []cache.InformerSynced{secrets.Informer().HasSynced},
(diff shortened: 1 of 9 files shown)


sanchezl:
/payload-job periodic-ci-openshift-hypershift-release-4.16-periodics-e2e-aws-ovn-conformance

soltysh (approved):
/approve

pkg/internalregistry/controllers/keyid_observation_controller.go
```diff
@@ -104,6 +110,38 @@ func (c *keyIDObservation) sync(ctx context.Context, key string) error {
 	return nil
 }
 
+func (c *keyIDObservation) fallbackKeyIDObservation(ctx context.Context) error {
+	_, err := c.secrets.Secrets("openshift-kube-apiserver").Get("bound-service-account-signing-key")
+	if err != nil && !errors.IsNotFound(err) {
+		// signing key secret exists, skip
+		return err
+	}
```
benluddy, line 118: Is it safe to make any assumptions one way or the other about the secret's existence based on this information?

pkg/internalregistry/controllers/keyid_observation_controller.go
```diff
@@ -104,6 +110,38 @@ func (c *keyIDObservation) sync(ctx context.Context, key string) error {
...
+
+	// create a throwaway API token
+	expirationSeconds := int64(10 * time.Minute / time.Second)
```
benluddy, line 121: Is this arbitrary? How short can the requested expiration time be?

pkg/internalregistry/controllers/keyid_observation_controller.go
```diff
@@ -112,6 +150,9 @@ func (c *keyIDObservation) Run(ctx context.Context, workers int) {
 	if !cache.WaitForNamedCacheSync(name, ctx.Done(), c.cacheSyncs...) {
 		return
 	}
+	if err := c.fallbackKeyIDObservation(ctx); err != nil {
+		runtime.HandleError(err)
+	}
```
benluddy, line 155: I see in the description "OCM would need to be restarted if the service account signing public key changes," but what if this one-shot attempt fails with a transient error? It sounds like it would not recover on its own.

Does it make sense to do the fallback as part of the controller sync if and only if the signing key secret does not exist?

pkg/internalregistry/controllers/keyid_observation_controller.go
sanchezl: We shouldn't reach here unless the cache has synced. If the secret should exist, but doesn't at this point, I don't see the harm in performing the fallback here, and the regular control loop will eventually pick up the secret and override if needed.

pkg/internalregistry/controllers/keyid_observation_controller.go
sanchezl: The controller sync is triggering on the one secret, a secret that wouldn't exisit to trigger the sync in hypershift, so I don't think I can add the fall back to the controller sync.

I can add a few retries on err.

pkg/internalregistry/controllers/keyid_observation_controller.go
benluddy: I think there are 3 cases:
1. The secret exists (`err == nil`) and the fallback is performed. The regular control loop should eventually catch up.
2. The secret does not exist (`errors.IsNotFound(err)`) and the fallback is performed. Even if the secret is later created, the regular control loop takes over.
3. Some different error occurred, and the fallback is _not_ performed. This seems like a problem on a cluster that will never have the signing key secret.

sanchezl: I see it now. Fixed.

pkg/internalregistry/controllers/keyid_observation_controller.go
```diff
@@ -0,0 +1,192 @@
...
+	err := wait.PollUntilContextTimeout(ctx, 30*time.Second, 10*time.Minute, true, func(ctx context.Context) (bool, error) {
+		_, err := c.secrets.Secrets("openshift-kube-apiserver").Get("bound-service-account-signing-key")
+		if err != nil && !errors.IsNotFound(err) {
+			// signing key secret exists, skip and let sync handle
+			return true, nil
+		}
+		if err != nil {
+			runtime.HandleError(err)
+			return false, nil
+		}
```
benluddy, line 123: Does this ever need to return true (and stop)?

```suggestion
		if err == nil {
			// signing key secret exists, skip and let sync handle
			return false, nil
		} else if !errors.IsNotFound(err) {
			// error other than notfound, signing key secret may or may not exist
			runtime.HandleError(err)
			return false, nil
		}
```

sanchezl: If the secret DNE, return `true` and stop the inner polling loop. `runWorker()` itself is running in an `UntilContext` loop.
If there is an error, the inner loop retries every 10 seconds for a minute.

benluddy:
/lgtm
/hold

@sanchezl Do you want to re-run the hypershift payload job?

benluddy:
/payload-job periodic-ci-openshift-hypershift-release-4.16-periodics-e2e-aws-ovn-conformance

PR #306, opened 2024-05-16
OCPBUGS-33815: openshift-controller-manager overwriting/undoing changes to ServiceAccount imagePullSecrets

sanchezl:
The ServiceAccount.ImagePullSecrets field does not have a merge strategy defined. As such, when using (server-side) Apply to update the ImagePullSecrets field, we have to be careful to not clobber the contents added by other clients. This is in contrast to the ServiceAccounts.Secret field, where ownership is at the level of an individual entry in the list.

### DIFF
--- a/pkg/internalregistry/controllers/service_account_controller.go
+++ b/pkg/internalregistry/controllers/service_account_controller.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"slices"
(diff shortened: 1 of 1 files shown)


soltysh (approved):
/lgtm
/approve

sjenning:
hypershift is experiencing a CI outage

/override ci/prow/e2e-hypershift-conformance

PR #318, opened 2024-07-02
OCPBUGS-35731: Race condition when deleting ServiceAccount

sanchezl:
- Altered the apply API calls to prevent the re-creation of a service account deleted between the beginning of a control loop and the API call. 
- Hardened apply API calls for the manage image pull secrets to prevent accidental re-creation.
- improved handling of deletion events

### DIFF
--- a/pkg/internalregistry/controllers/image_pull_secret_controller.go
+++ b/pkg/internalregistry/controllers/image_pull_secret_controller.go
@@ -147,6 +147,9 @@ func (c *imagePullSecretController) sync(ctx context.Context, key string) (error
 				return err, 0
 			}
 			patch.WithAnnotations(map[string]string{InternalRegistryAuthTokenTypeAnnotation: AuthTokenTypeBound})
+			// add the UID to the patch to ensure we don't re-create the secret if it no longer exists.
+			// the service account controller is responsible for re-creating the initial secret.
+			patch.WithUID(secret.UID)
 			_, err = c.client.CoreV1().Secrets(secret.Namespace).Apply(ctx, patch, metav1.ApplyOptions{Force: true, FieldManager: imagePullSecretControllerFieldManager})
 			if err != nil {
 				return err, 0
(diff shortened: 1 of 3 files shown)


pkg/internalregistry/controllers/legacy_token_secret_controller.go
```diff
@@ -72,12 +72,6 @@ func NewLegacyTokenSecretController(client kubernetes.Interface, secrets informe
 					c.queue.Add(key)
 				}
 			},
-			DeleteFunc: func(obj any) {
-				key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
-				if err == nil {
-					c.queue.Add(key)
-				}
-			},
```
sanchezl, line 80: Nothing for this controller to do on deletion.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -68,12 +68,6 @@ func NewServiceAccountController(kubeclient kubernetes.Interface, serviceAccount
 				c.queue.Add(key)
 			}
 		},
-		DeleteFunc: func(obj any) {
-			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
-			if err == nil {
-				c.queue.Add(key)
-			}
-		},
```
sanchezl, line 76: Nothing for this controller to do on deletion of a service account. Note that a deleted service account might still be queued by the Secret handler below.

pkg/internalregistry/controllers/image_pull_secret_controller.go
```diff
@@ -136,6 +136,9 @@ func (c *imagePullSecretController) sync(ctx context.Context, key string) (error
 	if err != nil {
 		return err, 0
 	}
+	if !secret.DeletionTimestamp.IsZero() {
+		return nil, 0
+	}
```
benluddy, line 141: If something prevents removal of the last finalizer on this secret, or many secrets, indefinitely, what would the systemic impact be? If any, would it be safer to remove this short-circuit even if that means the apply request could be pointless?

sanchezl: If I remove the short circuit, I can see the token being refreshed as a nice thing to keep things possibly going. But on the other hand, if the corresponding service account it already gone (a typical reason why the secret is being deleted), then we're not making any progress anyway.

dinhxuanvu (approved):
/lgtm

soltysh (approved):
/approve

dinhxuanvu:
/assign @jitendar-singh
PTAL

PR #319, opened 2024-07-10
OCPBUGS-36833: 4.16 "Bad" reconciliation loops can cause unbounded dockercfg secret creation

sanchezl:
* The image-registry service account controller attempts to recover a clobbered annotation before deciding to creating a new one.
* The image-registry image pull secret controller cleans up orphaned managed image pulled secrets (deletes the secret and removes references from the service account). Note that references will remain in the service account if the secrets are deleted before applying this patch.

PR #321, opened 2024-07-12
Revert "OCPBUGS-36833: 4.16 "Bad" reconciliation loops can cause unbounded dockercfg secret creation"

soltysh:
Reverts openshift/openshift-controller-manager#319

/cc @sanchezl

PR #322, opened 2024-07-12
Revert Revert OCPBUGS-36833: 4.16 "Bad" reconciliation loops can cause unbounded dockercfg secret creation

sanchezl:
Reapplying #319 with fix for nil pointer dereference panic

### DIFF
--- a/pkg/cmd/controller/imageregistry.go
+++ b/pkg/cmd/controller/imageregistry.go
@@ -12,7 +12,7 @@ func RunInternalImageRegistryPullSecretsController(ctx *ControllerContext) (bool
 	additionalRegistryURLs := ctx.OpenshiftControllerConfig.DockerPullSecret.RegistryURLs
 
 	serviceAccountController := controllers.NewServiceAccountController(kc, serviceAccounts, secrets)
-	imagePullSecretController, kids, urls := controllers.NewImagePullSecretController(kc, secrets)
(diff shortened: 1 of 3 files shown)


soltysh:
/hold
since we're thinking about merging the revert https://github.com/openshift/openshift-controller-manager/pull/321

sanchezl:
/payload periodic-ci-openshift-release-master-ci-4.17-e2e-gcp-ovn-upgrade

sanchezl:
/hold cancel 
payload jobs completed successfully

vrutkovs (approved):
/lgtm

pgodowski:
@sanchezl if I correctly understand the PR, it will address the issue where the image pull Secrets were generated in the massive scale, at the same time pruning the previously created image pull Secrets which are now orphaned. Good.

But seems this PR does not resolve the scenario where there is 3rd party controller creating / managing ServiceAccount and where the annotation `InternalRegistryAuthTokenServiceAccountAnnotation` is removed by the 3rd party controller and then re-added by OCP SA controller, right? Meaning, that there will be still some form of "race condition"  between two controllers, but at least there will be no side effect of it, resulting with thousands of Secrets created?

If so, what is the plan / issue to resolve that "race condition" scenario? The net results is that 3rd party controllers will start spinning around, together with OCP, consuming compute resources without any good reason.

I think that the proper fix is that the SA controller should create the image pull secret and then annotate such Secret with the reference to its SA, rather than annotate SA to point to the Secret.

PR #325, opened 2024-07-25
Jitter expiry of image pull secret token to smooth refresh load.

benluddy:


pkg/internalregistry/controllers/image_pull_secret_controller.go
```diff
@@ -175,11 +176,17 @@ func (c *imagePullSecretController) sync(ctx context.Context, key string) (error
 	var serviceAccountName = serviceAccountNameForManagedSecret(secret)
 	klog.V(2).InfoS("Refreshing image pull secret", "ns", secret.Namespace, "name", secret.Name, "serviceaccount", serviceAccountName)
 
-	// request new token, bound by default time and bound to this secret
+	// request new token, bound to this secret and with jittered expiration
+	expirationSeconds := 60*60 + rand.Int64N(10*60) // [1h, 1h10m)
```
sanchezl, line 180: The default ExpirationSeconds is configurable, so we can't just assume it is 1h.

sanchezl: > ~~The default ExpirationSeconds is configurable, so we can't just assume it is 1h.~~

I was wrong about that, only the max ExpirationSeconds is configuration.

PR #337, opened 2024-09-17
OCPBUGS-42097: user system:serviceaccount:openshift-infra:serviceaccount-pull-secrets-controller in ns/openshift-infra must not produce too many applies

sanchezl:
An Apply request was being made unnecessarily against secrets that did not require any updates.

### DIFF
--- a/pkg/internalregistry/controllers/service_account_controller.go
+++ b/pkg/internalregistry/controllers/service_account_controller.go
@@ -3,10 +3,12 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"slices"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -134,16 +136,47 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
 	// name of managed secret
 	secretName, err := c.managedImagePullSecretName(ctx, serviceAccount)
 
-	// ensure secret ref annotation is set
-	if serviceAccount.Annotations[InternalRegistryImagePullSecretRefKey] != secretName {
-		patch, err := applycorev1.ExtractServiceAccount(serviceAccount, serviceAccountControllerFieldManager)
-		if err != nil {
-			return err
+	patch := applycorev1.ServiceAccount(name, ns).
+		WithAnnotations(map[string]string{InternalRegistryImagePullSecretRefKey: secretName}).
+		// TODO stop adding to .Secrets part of API-1798
+		WithSecrets(applycorev1.ObjectReference().WithName(secretName))
+
+	// Extract an apply patch from the service account. The extracted patch is used
+	// to deduce some field ownership information without having to parse
+	// ManagedFields, and it must remain unmodified so we can detect if the apply
+	// call is necessary.
+	extracted, err := applycorev1.ExtractServiceAccount(serviceAccount, serviceAccountControllerFieldManager)
+	if err != nil {
+		return err
+	}
+
+	// As an atomic field, an update to ImagePullSecrets replaces the the entire
+	// field. We want to be cautious about not overwriting the changes made to
+	// ImagePullSecrets by other users or controllers. When updating ImagePullSecrets
+	// always add the ResourceVersion to the patch to ensure the apply fails if the
+	// cache was stale.
+	if slices.ContainsFunc(serviceAccount.ImagePullSecrets, func(ref corev1.LocalObjectReference) bool { return ref.Name == secretName }) {
+		// If this controller owns the field, we copy the value from the extracted patch
+		// to preserve it upon apply with force=true. If this controller does not own the
+		// field, the copied extracted value should be nil, also preserving the existing
+		// value upon apply with force=true.
+		patch.ImagePullSecrets = extracted.ImagePullSecrets
+	} else {
+		// preserve the existing ImagePullSecrets items and add the managed image pull secret.
+		for _, ref := range serviceAccount.ImagePullSecrets {
+			patch.WithImagePullSecrets(applycorev1.LocalObjectReference().WithName(ref.Name))
 		}
-		patch.WithAnnotations(map[string]string{InternalRegistryImagePullSecretRefKey: secretName})
+		patch.WithImagePullSecrets(applycorev1.LocalObjectReference().WithName(secretName))
+	}
+
+	// apply patch only if necessary
+	if !equality.Semantic.DeepEqual(patch, extracted) {
 		// add the UID to the patch to ensure we don't re-create the service account if it no longer exists
 		patch.WithUID(serviceAccount.UID)
-		// we apply this now to ensure the secret name stays stable in case a error occurs while reconciling
+		if len(patch.ImagePullSecrets) > 0 {
+			// prevent inadvertently overwriting someone else's updates to ImagePullSecrets
+			patch.WithResourceVersion(serviceAccount.ResourceVersion)
+		}
 		serviceAccount, err = c.client.CoreV1().ServiceAccounts(ns).Apply(ctx, patch, metav1.ApplyOptions{Force: true, FieldManager: serviceAccountControllerFieldManager})
 		if err != nil {
 			return err
@@ -165,9 +198,10 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
 	var secretOwnerRefNeedsUpdate, secretSARefNeedsUpdate bool
 	if secret != nil {
 		secretSARefNeedsUpdate = secret.Annotations[InternalRegistryAuthTokenServiceAccountAnnotation] != serviceAccount.Name
+		secretOwnerRefNeedsUpdate = true
 		for _, ref := range secret.OwnerReferences {
 			if ref.Name == serviceAccount.Name && ref.UID == serviceAccount.UID && ref.Kind == "ServiceAccount" && ref.APIVersion == "v1" {
-				secretOwnerRefNeedsUpdate = true
+				secretOwnerRefNeedsUpdate = false
 				break
 			}
 		}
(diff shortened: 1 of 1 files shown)


pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -226,6 +228,15 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
 	}
 
+	// skip apply if patch seems to be a no-op
+	extractedPatch, err := applycorev1.ExtractServiceAccount(serviceAccount, serviceAccountControllerFieldManager)
+	if err != nil {
+		return err
+	}
+	if equality.Semantic.DeepEqual(extractedPatch, patch) {
+		return nil
+	}
```
benluddy, line 238: Is this the right comparison between two apply configurations? As I understand it, a nil field value in an apply configuration means "this manager has no opinion on this field" whereas a non-nil but empty field value means "this manager wants this field to have the empty value". The special case comparisons in Semantic (like comparing metav1.Time with normalized time zone) are important, though. The best bet might be to marshal both and compare bytes?

sanchezl: > a nil field value in an apply configuration means "this manager has no opinion on this field" 
Unless, the manager owns that field, in which case it means "I no longer want to provide a value for this field, remove me as the owner and delete it if it has no other owners."

benluddy: Potato, potato? :smile:

sanchezl: I've instrumented a local copy and it seems to always return what I expect. I can't think of a scenario where this would go wrong.

benluddy: Oh, all the slice fields in generated ApplyConfiguration types have omitempty nil and empty slices both serialize to the same thing... is it possible to express the intent "this atomic list should be empty" with SSA?

sanchezl: With SSA it seems you can't explicitly say "this list should be empty", only "I have nothing to add to this list".

It is unfortunate that this list is atomic, as it means that there is a chance that two competing controllers might experience some contention for the final result, but unless they are misbehaving (e.g. inconsistent sort) or deadlocked on each other, they should eventually reconcile.

I could remove the ImagePullSecret update out of the Apply call and invoke a separate Update just for that field, but is it worth the extra complexity?

benluddy: I think that question can be dealt with separately from skipping the no-op apply requests, especially since it already exists. It might be true in practice that there are few or no clients whose writes contend with this controller (I don't know) but if there are, it would be difficult to argue that those clients are misbehaving if they don't recover from it.

sanchezl:
/override ci/prow/e2e-gcp-ovn-builds

/override ci/prow/openshift-e2e-aws-ovn-builds-techpreview

Known issue related to ruby imagestream bump: https://issues.redhat.com/browse/OCPBUGS-42237

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -134,16 +139,56 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
+
+	// we always want the annotation set
+	patch.WithAnnotations(map[string]string{InternalRegistryImagePullSecretRefKey: secretName})
+
+	// TODO stop adding to .Secrets part of API-1798
+	if !slices.ContainsFunc(patch.Secrets, func(ref applycorev1.ObjectReferenceApplyConfiguration) bool { return *ref.Name == secretName }) {
+		patch.WithSecrets(applycorev1.ObjectReference().WithName(secretName))
+	}
+
+	// We have tp be careful with the atomic ServiceAccount.ImagePullSecrets field so
```
benluddy, line 156:
```suggestion
	// We have to be careful with the atomic ServiceAccount.ImagePullSecrets field so
```

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -134,16 +139,56 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
-		if err != nil {
-			return err
+	// extract the apply configuration patch from the cache
+	patch, err := applycorev1.ExtractServiceAccount(serviceAccount, serviceAccountControllerFieldManager)
+	if err != nil {
+		return err
+	}
+
+	// we always want the annotation set
+	patch.WithAnnotations(map[string]string{InternalRegistryImagePullSecretRefKey: secretName})
```
benluddy, line 149: I think we want to "assume an open world" and build the desired apply configuration from scratch rather than by modifying the existing apply configuration. As an added benefit, that should remove the need for some of the conditional branches in the control loop, too.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -134,16 +139,56 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
-		patch.WithAnnotations(map[string]string{InternalRegistryImagePullSecretRefKey: secretName})
+		patch.WithImagePullSecrets(applycorev1.LocalObjectReference().WithName(secretName))
+	}
+
+	// apply patch only if necessary
+	extracted, err := applycorev1.ExtractServiceAccount(serviceAccount, serviceAccountControllerFieldManager)
+	if err != nil {
+		return err
+	}
+	if !cmp.Equal(patch, extracted) {
```
benluddy, line 184: After our earlier discussion, and because the all of the generated ApplyConfiguration fields have omitempty in their JSON tag, I think we want to use `apiequality.Semantic` for this comparison.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -134,16 +139,56 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
 		// add the UID to the patch to ensure we don't re-create the service account if it no longer exists
 		patch.WithUID(serviceAccount.UID)
-		// we apply this now to ensure the secret name stays stable in case a error occurs while reconciling
+		// if setting the atomic ImagePullSecrets field, add the resourceVersion to the
+		// patch to ensure we do not race with another operator or user.
+		if len(patch.ImagePullSecrets) > 0 {
+			patch.WithResourceVersion(serviceAccount.ResourceVersion)
+		}
```
benluddy, line 191: If we're building the desired configuration from scratch, then ResourceVersion could be set at the same time ImagePullSecrets.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -165,9 +210,10 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
 	var secretOwnerRefNeedsUpdate, secretSARefNeedsUpdate bool
```
benluddy, line 210: Does this also become simpler if the desired apply configuration is built and compared against the existing apply configuration, like you're doing with the ServiceAccount apply, instead of looking for specific fields that are out of level?

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -195,61 +241,7 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
 		}
```
benluddy, line 241: GitHub won't let me comment on L236 because you aren't touching it here, but is the update to Data vulnerable to the same issue as with merging the observed and desired entries of ImagePullSecrets on the ServiceAccount? Here the apply configuration is setting the docker config key to a value based on the value in the informer cache, and if the cache is stale it will stomp a recent write.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -134,16 +139,56 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
+	// we always want the annotation set
+	patch.WithAnnotations(map[string]string{InternalRegistryImagePullSecretRefKey: secretName})
+
+	// TODO stop adding to .Secrets part of API-1798
+	if !slices.ContainsFunc(patch.Secrets, func(ref applycorev1.ObjectReferenceApplyConfiguration) bool { return *ref.Name == secretName }) {
+		patch.WithSecrets(applycorev1.ObjectReference().WithName(secretName))
+	}
+
+	// We have tp be careful with the atomic ServiceAccount.ImagePullSecrets field so
+	// that we do not race with another controller or user. Add ImagePullSecrets to
```
benluddy, line 157:
```suggestion
	// that we do not contend with another controller or user over field ownership. Add ImagePullSecrets to
```

Would that be more precise? Basically, this controller expects that some ServiceAccounts will have others writing to the atomic ImagePullSecrets list, but as long as other writers preserve the specific entry it cares about, it won't insist that it is also the field's owner.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: I do believe that that we can "simplify" in that manner, but I was trying to limit the changes in this PR to the serviceAccount updates.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: Reworded.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: Done.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: Reworded.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: I purposely choose to extracted the patch. Comparing the extracted `ImagePullSecrets` to the cached `ImagePullSecrets` allows me to detect the case where we do not own the field, yet there is no need to update the `ImagePullSecrets`.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -117,6 +119,8 @@ func NewServiceAccountController(kubeclient kubernetes.Interface, serviceAccount
 
 const serviceAccountControllerFieldManager = "openshift.io/image-registry-pull-secrets_service-account-controller"
 
+var imagePullSecretFieldPath, _ = fieldpath.MakePath("imagePullSecrets")
```
benluddy, line 122: No longer needed in the latest revision?

sanchezl: Removed.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: I refactored to start with a empty patch instead of an extracted one, but it still make sense to me to add the RV after the desired/extracted patch comparison.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: Refactored.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -134,16 +136,47 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
...
+
+	// As an atomic field, an update to ImagePullSecrets replaces the the entire
+	// field. We want to be cautious about not overwriting the changes made to
+	// ImagePullSecrets by other users or controllers. When updating ImagePullSecrets
+	// always add the ResourceVersion to the patch to ensure the apply fails if the
+	// cache was stale.
+	if slices.ContainsFunc(serviceAccount.ImagePullSecrets, func(ref corev1.LocalObjectReference) bool { return ref.Name == secretName }) {
+		// If this controller owns the field, we copy the value from the extracted patch
+		// to preserve it upon apply with force=true. If this controller does not own the
+		// field, the copied extracted value should be nil, also preserving the exiting
```
benluddy, line 161:
```suggestion
		// field, the copied extracted value should be nil, also preserving the existing
```

benluddy:
This looks right to me. I'm assuming there is reasonable existing test coverage and that we've observed the expected reduction in applies somewhere (placing a hold in case you're still checking this for the latest revision).

/lgtm
/hold

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -165,9 +198,10 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
 	var secretOwnerRefNeedsUpdate, secretSARefNeedsUpdate bool
 	if secret != nil {
 		secretSARefNeedsUpdate = secret.Annotations[InternalRegistryAuthTokenServiceAccountAnnotation] != serviceAccount.Name
+		secretOwnerRefNeedsUpdate = true
 		for _, ref := range secret.OwnerReferences {
 			if ref.Name == serviceAccount.Name && ref.UID == serviceAccount.UID && ref.Kind == "ServiceAccount" && ref.APIVersion == "v1" {
-				secretOwnerRefNeedsUpdate = true
+				secretOwnerRefNeedsUpdate = false
```
adambkaplan, line 204: Why are we flipping this from `true` to `false`? This inverts the logic of the for loop, while keeping the break condition unchanged.

pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -195,61 +229,7 @@ func (c *serviceAccountController) sync(ctx context.Context, key string) error {
 		}
 	}
 
-	// nothing else to do if we are not dealing with a bound image pull secret
-	if secret.Annotations[InternalRegistryAuthTokenTypeAnnotation] != AuthTokenTypeBound {
-		return nil
-	}
```
adambkaplan, line 201: I don't see this logic captured above - was this removal intentional?

adambkaplan (commented):
I have a few questions below. I'm also concerned there are no corresponding unit test updates, which I think will help clarify the desired behavior (and definitely catch potential bugs/regressions in the future).

cc @flavianmissi - Image Registry team should review this as well.

flavianmissi:
I'm not that familiar with this code but as far as I can see everything looks solid, thanks for doing this @sanchezl 🙏🏼

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: Yes, this was intentional.

pkg/internalregistry/controllers/service_account_controller.go
sanchezl: Logic was previously inverted. What we are trying to say here is: If the secret contains the service account as an owner already, we do not need to update the secret.

PR #365, opened 2025-02-11
OCPBUGS-50507: Intermittent authentication issues when accessing OpenShift registry

sanchezl:
In some instances, the controller would re-queue an image pull secret in such a way that the check for an expiring token would only occur after the token had expired. This would result in a gap of time in which the pull secret contents would be invalid.

### DIFF
--- a/pkg/internalregistry/controllers/image_pull_secret_controller.go
+++ b/pkg/internalregistry/controllers/image_pull_secret_controller.go
@@ -7,9 +7,9 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+	"sync/atomic"
 	"time"
 
-	"go.uber.org/atomic"
 	"golang.org/x/exp/slices"
 	"gopkg.in/go-jose/go-jose.v2/jwt"
 	authenticationv1 "k8s.io/api/authentication/v1"
@@ -52,8 +52,8 @@ func NewImagePullSecretController(kubeclient kubernetes.Interface, secrets infor
 		serviceAccounts: serviceAccounts.Lister(),
 		cacheSyncs:      []cache.InformerSynced{secrets.Informer().HasSynced, serviceAccounts.Informer().HasSynced},
 		queue:           workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "bound-token-managed-image-pull-secrets"),
-		kids:            atomic.NewPointer[[]string](nil),
-		urls:            atomic.NewPointer[[]string](nil),
+		kids:            &atomic.Pointer[[]string]{},
+		urls:            &atomic.Pointer[[]string]{},
 		kidsC:           make(chan []string),
 		urlsC:           make(chan []string),
 	}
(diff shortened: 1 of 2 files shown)


pkg/internalregistry/controllers/image_pull_secret_controller.go
```diff
@@ -329,29 +330,35 @@ CheckUrl:
...
 	}
 	klog.V(4).InfoS("Internal registry pull secret does not need to be refreshed.", "ns", imagePullSecret.Namespace, "name", imagePullSecret.Name)
-	return true, requeueAt
+	return true, &requeueAt
+}
+
+func refreshThresholdTime(nbf, exp time.Time) time.Time {
+	// calculate the time at which only 40% of the valid duration would be left
+	validDuration := exp.Sub(nbf)
+	return exp.Add(-time.Duration(int64(float64(validDuration) * 0.4)))
```
ricardomaraschini, line 361: Are we guarantee to always get `exp` > `nbf` here ?

pkg/internalregistry/controllers/image_pull_secret_controller.go
```diff
@@ -267,31 +268,31 @@ func dockerConfig(token string, urls []string) any {
...
 
-func (c *imagePullSecretController) isSecretRefreshNeeded(secret *corev1.Secret, urls, kids []string) (bool, time.Time) {
-	valid, refreshAt := c.registryAuthenticationFileValid(secret, urls, kids)
+func (c *imagePullSecretController) isSecretRefreshNeeded(secret *corev1.Secret, urls, kids []string, now time.Time) (bool, *time.Time) {
+	valid, refreshAt := c.registryAuthenticationFileValid(secret, urls, kids, now)
 	return !valid, refreshAt
 }
 
-func (c *imagePullSecretController) registryAuthenticationFileValid(imagePullSecret *corev1.Secret, urls, kids []string) (bool, time.Time) {
+func (c *imagePullSecretController) registryAuthenticationFileValid(imagePullSecret *corev1.Secret, urls, kids []string, now time.Time) (bool, *time.Time) {
```
ricardomaraschini, line 276: Wouldn't it be a good idea moving parts of this function to `isSecretRefreshNeeded()` instead ? I mean, this function does some validations (if the secret is of the right `Type` for instance) while also validates if the tokens in a valid secret haven't expired. I see this as two distinct things: in one the secret is invalid while in the other the tokens are invalid. It is up to you but I believe that splitting would make things easier to reason about.

The way things are here the only purpose of `isSecretRefreshNeeded()` is to invert the value of `valid`.

ricardomaraschini (commented):
This seems to move us to a better place! Thanks for that. I would like to see some unit tests for the `registryAuthenticationFileValid()` and `refreshThresholdTime()`, if that is possible.

pkg/internalregistry/controllers/image_pull_secret_controller.go
droslean: Can we have some unit tests to test the logic as well?

pkg/internalregistry/controllers/image_pull_secret_controller.go
sanchezl: It would be very unlikely, it means k8s would have issued an invalid token. I added a check anyway.

PR #381, opened 2025-05-11
WRKLDS-1676: Delete unused pkg/serviceaccounts

tchap:
Once the unused function RunServiceAccountPullSecretsController
is removed and a few constants pulled out of the package,
it is actually unused and can be deleted.


pkg/internalregistry/controllers/service_account_controller.go
```diff
@@ -304,10 +311,10 @@ func (c *serviceAccountController) legacyImagePullSecretName(ctx context.Context
...
 
 var expectedLegacyAnnotations = map[string]func(*corev1.ServiceAccount, string) bool{
-	corev1.ServiceAccountNameKey:                   func(sa *corev1.ServiceAccount, v string) bool { return sa.Name == v },
-	corev1.ServiceAccountUIDKey:                    func(sa *corev1.ServiceAccount, v string) bool { return sa.UID == types.UID(v) },
-	controllers.ServiceAccountTokenSecretNameKey:   func(sa *corev1.ServiceAccount, v string) bool { return true },
-	controllers.ServiceAccountTokenValueAnnotation: func(sa *corev1.ServiceAccount, v string) bool { return true },
+	corev1.ServiceAccountNameKey:       func(sa *corev1.ServiceAccount, v string) bool { return sa.Name == v },
+	corev1.ServiceAccountUIDKey:        func(sa *corev1.ServiceAccount, v string) bool { return sa.UID == types.UID(v) },
+	ServiceAccountTokenSecretNameKey:   func(sa *corev1.ServiceAccount, v string) bool { return true },
+	ServiceAccountTokenValueAnnotation: func(sa *corev1.ServiceAccount, v string) bool { return true },
```
sanchezl, line 317: Let's use these strings directly:

```suggestion
	"openshift.io/token-secret.name":   func(sa *corev1.ServiceAccount, v string) bool { return true },
	"openshift.io/token-secret.value": func(sa *corev1.ServiceAccount, v string) bool { return true },
```

sanchezl (changes requested):

PR #383, opened 2025-05-20
OCPBUGS-56474: Empty proxy variables are causing issues during the build

kunalmemane:
Problem:
Irrespective of default proxy settings these proxy variables are set on build containers. If any defaults are set they reflect those setting ,if not set they are set to null on the build container. Having these null variables are causing problems to some customers as its breaking some required apps on container HTTP_PROXY,http_proxy,HTTPS_PROXY,https_proxy,NO_PROXY,no_proxy

Fix:
Code changed only to add these proxy variables only if they are defined in defaults and if they are not null.


de1987 (approved):
verified, lgtm!

PR #387, opened 2025-06-03
OCPBUGS-57198: Make the pull secrets volume mounts in build containers read-only.

benluddy:


prabhapa:
@benluddy pls replace the tracker name with this in the PR name OCPBUGS-57198

benluddy:
/retitle OCPBUGS-57198: Make the pull secrets volume mounts in build containers read-only.

PR #397, opened 2025-06-20
OCPBUGS-57918: ignore error failing to find pull/push secrets

dusk125:
Ignoring the error returned when we don't have a push/pull secret and are attempting to find one in the service accounts. This should fix issues with using public registries with no secrets defined in the config, and rely on the image puller to error out if not defining a secret and attempting to pull/push from a private registry.

/cherry-pick 4.19
/cherry-pick 4.18
/cherry-pick 4.17
/assign @sanchezl


sanchezl (changes requested):
Let's fix the underlying funcs such that:
* Errors getting the SA bubble up
* Errors getting secrets (except for NotFound) bubble up
* In keeping with your concept of letting the underlying puller to error out, a null reference should be a valid response.

sanchezl (approved):
/lgtm

PR #447, opened 2026-06-08
OCPBUGS-77404: bugfix: use a system context for setting custom auth file path when fetching image sources

everettraven:
so that the global pull secret can be mounted to the openshift-controller-manager pod as a volume and used when attempting to fetch image signatures for import.

### DIFF
--- a/pkg/image/controller/signature/container_image_downloader.go
+++ b/pkg/image/controller/signature/container_image_downloader.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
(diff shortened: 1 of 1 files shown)


ehearne-redhat:
After going through the other PR mentioned here, I can understand the need for this change. Since we are proposing to mount `/etc/containers` this change is needed for that. From the mentioned bug ticket, I can see at present that the controller-manager makes requests to quay.io and not the internal image registry. It seems it doesn't have access to that registry and bringing in this change will enable that as the config.json is read in this way. 

/lgtm

